PR #25, opened 2015-10-21
Always use absolute urls

hxoliverrumbelow:
Stop responding with relative URLs:
```
links: {
    self: "/rest/articles/de305d54-75b4-431b-adb2-eb6b9e546014?include=author"
},
```
and start using absolute URLs:
```
links: {
    self: "http://localhost:16006/rest/articles/de305d54-75b4-431b-adb2-eb6b9e546014?include=author"
},
```

I'm not 100% happy with this yet, but its a step in the right direction. I'm thinking about if we can decouple the domain name from the config and simply use whatever gets passed in per-request. It'd also be nice to run an API on a different port from what the customer see's to make better use of infrastructure. This stuff will become clearer in the next few weeks when we look at pushing this into production.

### DIFF
--- a/lib/jsonApi.js
+++ b/lib/jsonApi.js
@@ -9,6 +9,7 @@ var router = require("./router.js");
 var responseHelper = require("./responseHelper.js");
 var routes = require("./routes");
 var mockHandlers = require("./mockHandlers.js");
+var url = require("url");
 
 jsonApi.Joi = ourJoi.Joi;
 jsonApi.mockHandlers = mockHandlers.handlers;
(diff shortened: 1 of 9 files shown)


hxoliverrumbelow:
Addresses #25

lib/jsonApi.js
```diff
@@ -16,12 +16,22 @@ jsonApi.mockHandlers = mockHandlers.handlers;
...
 
 jsonApi.authenticate = router.authenticateWith;
 
+jsonApi._concatenateUrlPrefix = function(config) {
+  var pathPrefix = "";
+  pathPrefix += (config.protocol || "https") + "://";
+  pathPrefix += (config.hostname || "localhost");
+  if (config.port) pathPrefix += ":" + config.port;
+  pathPrefix += config.base;
+  return pathPrefix;
```
pmcnr-hx: Would it be worth using the standard library [`url`](https://nodejs.org/api/url.html) module here to build a url object and then format it rather than concatenating it manually?

pmcnr-hx:
Step in the right direction: :+1:

lib/jsonApi.js
hxoliverrumbelow: Good call :+1:

PR #26, opened 2015-11-01
Windows development

richardschneider:
I'm using a windows development environment and ran into the following issues

- some the filenames are not valid in windows land
- some of the scripts in `package.json` did not run


hxoliverrumbelow:
Hey @richardschneider - Thanks for the contribution :+1: It looks like your two PR's have gotten a bit mixed up, they both contain the same combined changeset. Could you separate them out?

richardschneider:
Sorry about that.  I'm new to pull requests.  My mistake was making the 1st PR from `master` and not an `issues` branch.  This then pulled in 2nd PR when merged into `master`.

Not sure how to fix this?

example/resources/tags.js
```diff
@@ -4,7 +4,7 @@ jsonApi.define({
   namespace: "json:api",
   resource: "tags",
   description: "Used to group resources together, useful for finding related resources.",
-  handlers: jsonApi.mockHandlers,
+//  handlers: jsonApi.mockHandlers,
```
hxoliverrumbelow: Can we put this code back in, the same for the others above? These are part of the example implementation provided with jsonapi-server, as such we need the example implementation to be as obvious as possible around how it behaves. Having a default fallback handler is fine, but it should be obvious to newcomers that they'll want to find/provide a more serious way of storing data.

lib/jsonApi.js
```diff
@@ -37,6 +38,7 @@ jsonApi._cleanBaseUrl = function(base) {
 
 jsonApi.define = function(resourceConfig) {
   resourceConfig.namespace = resourceConfig.namespace || "default";
+  resourceConfig.handlers = resourceConfig.handlers || jsonApi.defaultHandlers;
```
hxoliverrumbelow: Can we drop the addition on line 15 and tweak this line to simply `||` against `mockHandlers.handlers`? I can't see any future use for duplicating the property above?

hxoliverrumbelow:
I've added a few inline comments - if we can resolve them I'm happy to merge this as-is. Talking you through separating the two pieces of work would only prove frustrating for the both of us ;) 

I'd recommend trying something like this to work through git branching: http://pcottle.github.io/learnGitBranching/
or maybe this (the nodejs workshopper modules are usually fantastic)
https://www.npmjs.com/package/git-it

example/resources/tags.js
richardschneider: Well it makes it harder for the test of changing the default handler.  That's why a "commented out" the code.  But I do see your point.  Will try to update the patch sometime tomorrow my time (I'm in NZ +13:00 UTC).

lib/jsonApi.js
richardschneider: No worries.

example/resources/tags.js
hxoliverrumbelow: Hmm you're right. How about we compromise, lets take the handler out completely for just one of the resources and fully replace the line with a comment pointing out that it's opting in to the default mockHandlers?

pmcnr-hx:
Closing in favour of #41.

PR #38, opened 2015-11-12
Ready property and check for handlers

pmcnr-hx:
Handlers will probably take a bit to set-up and be ready to process requests. Also they might be temporarily unavailable. jsonapi-server should handle these cases gracefully.

### DIFF
--- a/lib/routes/helper.js
+++ b/lib/routes/helper.js
@@ -48,6 +48,15 @@ helper.verifyRequest = function(request, resourceConfig, res, handlerRequest, ca
     });
   }
 
+  if (!resourceConfig.handlers.ready) {
+    return helper.handleError(request, res, {
+      status: "503",
+      code: "EUNAVAILABLE",
+      title: "Resource not available",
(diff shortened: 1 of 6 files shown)


DuncanFenning:
Makes sense :+1:

lib/routes/helper.js
```diff
@@ -48,6 +48,15 @@ helper.verifyRequest = function(request, resourceConfig, res, handlerRequest, ca
     });
   }
 
+  if (!resourceConfig.handlers.ready) {
+    return helper.handleError(request, res, {
+      status: "503",
+      code: "EUNAVAILABLE",
+      title: "Resource not available",
```
hxoliverrumbelow: Is this the right title here? Could it be more like `Resource temporarily unavailable` - something more fitting to the detail provided below?

pmcnr-hx:
@hxoliverrumbelow: do the above changes address your concerns?

hxoliverrumbelow:
This looks glorious. If you can bump the version and add to the changelog this is good to merge :+1:

PR #46, opened 2015-11-19
Added a use function to enable the mounting of custom middleware.

msaspence:
Work from pair programming Thurs afternoon

### DIFF
--- a/example/server.js
+++ b/example/server.js
@@ -43,3 +43,4 @@ jsonApi.onUncaughtException(function(request, error) {
 jsonApi.start();
 server.start = jsonApi.start;
 server.close = jsonApi.close;
+server.use = jsonApi.use;
--- a/example/swagger-generator.js
+++ b/example/swagger-generator.js
@@ -1,0 +1,15 @@
+var swagger = module.exports = {};
+
+swagger.config = {
+  verb: "get",
+  path: "swagger.json",
+  skipDefaultHeaders: true
+};
+
+swagger.handler = function(request, resourceConfig, res) {
+  res.set({
+    "Content-Type": "text/html"
+  });
+  res.end("<strong>Hello World!</strong>");
+};
+
--- a/example/swagger-generator.js
+++ b/example/swagger-generator.js
@@ -1,3 +1,4 @@
+"use strict";
 var swagger = module.exports = {};
 
 swagger.config = {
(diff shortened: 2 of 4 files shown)


msaspence:
@hxoliverrumbelow I was thinking about this on the train home, it's just an idea not sure what the implementation might look like but...

Since you are already using Express as a base, what if `jsonApi.use` simply forwarded to `express.use`? 
This would present a few **possible** benefits:

 * Limits the amount of code you need to implement and test as you are utilizing functionality already available in express.
 * Presents an established and potentially familiar interface to 3rd party developers writing middleware
 * Allows any existing middleware that is available for Express to be used with jsonApi

hxoliverrumbelow:
It's an interesting one. In the past, to get things working, I've simply exposed the express server via a getter, so something like this:
```
var jsonApi = require("../.");
jsonApi.setConfig({ });
jsonApi.getExpress().use(/* stuff */);
jsonApi.start();
```

One thing I'm cautious about is making this project too dependant on ExpressJS. Right now, the only coupling we have with it is contained within `/lib/router.js` - only one small-ish file stands between us and different routing framework. If we come to optimise some of the more complex functionality such as the fetching of included documents we might find a different routing framework solves some of those challenges. In the past some of our devs have used the HapiJS routing framework to deal with resource inclusion. If we allow other modules to depend on ExpressJS, we won't have that flexibility to move to a different framework. The current solution gives us the freedom to jump frameworks.

Some of the use cases we should aim to handle are:

 * Access logs
 * Dynamic documentation
 * API discovery documents
 * robots.txt (?)
 * Web sockets - pushing and pulling documents via a socket instead of over HTTP https://www.npmjs.com/package/socket.io

Of those use cases, the web sockets scenario is the least trivial - there's a nice module out there that hooks up socket.io to express whose use case looks like this:
```
var app = require('express')();
var server = require('http').createServer(app);
var io = require('socket.io')(server);
io.on('connection', function(){ /* … */ });
server.listen(3000);
```
This use case won't work with either solution so far. It's probably worth thinking a bit harder about how this interface should work.

hxoliverrumbelow:
I've had a chat with @msaspence, we're going to defer this decision for now and come back to it at a later date.

seawatts:
Bummer, this is the only thing preventing me from using this library. Other than the the library looks great =)

theninj4:
@seawatts do you mind me asking what your use case looks like? We're migrating everything behind this module at work - to enable us to gracefully move things around we've got `jsonapi-server` sharing the same `express` instance as our core project. I can share how we've done it for now if that'll help? I've been thinking about documenting it for a while now...

PR #47, opened 2015-11-24
Auto Swagger Documentation

hxoliverrumbelow:
If, when setting the config for an api, the `swagger` property is set, a `swagger.json` spec will be generated and make available at `/swagger.json`.

For eample, the example server's swagger.json is available at http://localhost:16006/rest/swagger.json

### DIFF
--- a/lib/swagger/index.js
+++ b/lib/swagger/index.js
@@ -1,0 +1,74 @@
+"use strict";
+var swagger = module.exports = { };
+var jsonApi = require("../../");
+var swaggerPaths = require("./paths.js");
+var swaggerResources = require("./resources.js");
+
+swagger.generateDocumentation = function() {
+  var swaggerDoc = swagger._getSwaggerBase();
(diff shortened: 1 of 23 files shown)


hxoliverrumbelow:
The generated swagger file is tested for correctness against all sensible requests and responses made in the test suite.

lib/swagger/index.js
```diff
@@ -0,0 +1,74 @@
...
+  return {
+    swagger: "2.0",
+    info: {
+      title: swaggerConfig.title,
+      version: swaggerConfig.version,
+      description: swaggerConfig.description,
+      contact: {
+        name: (swaggerConfig.contact || { }).name,
+        email: (swaggerConfig.contact || { }).email,
+        url: (swaggerConfig.contact || { }).ur
```
DuncanFenning: I'm guessing this should be 'url'

pmcnr-hx:
Excellent effort! All looks reasonable: :+1:

PR #49, opened 2015-11-25
Callback with explicit value.

pmcnr-hx:
Callback with an explicit `null` for objects that failed validation to avoid breaking the `async.waterfall` where function is used because of a shorter number of arguments than expected.

* [ ] :cake: 
* [x] :pizza:


hxoliverrumbelow:
:+1:

PR #50, opened 2015-11-25
Deterministic order for arrays in tests.

pmcnr-hx:
If ordering is not specified, results may be in any order: make sure when we do comparisons the results are sorted.


DuncanFenning:
Nice :+1:

PR #52, opened 2015-11-25
Additional debugging output

hxoliverrumbelow:
I've added additional debugging opt-ins:

### jsonApi:validation:input
All models going through input validation

### jsonApi:validation:output
All models going through output validation

### jsonApi:include
All inclusion requests and any errors they give

### jsonApi:errors
All Error objects being sent out to clients

### DIFF
--- a/lib/postProcessing/include.js
+++ b/lib/postProcessing/include.js
@@ -7,6 +7,7 @@ var externalRequest = require("request").defaults({
   pool: { maxSockets: Infinity }
 });
 var async = require("async");
+var includeDebug = require("debug")("jsonApi:include");
 
 includePP.action = function(request, response, callback) {
   var includes = request.params.include;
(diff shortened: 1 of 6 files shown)


lib/postProcessing/include.js
```diff
@@ -7,6 +7,7 @@ var externalRequest = require("request").defaults({
   pool: { maxSockets: Infinity }
 });
 var async = require("async");
+var includeDebug = require("debug")("jsonApi:include");
```
pmcnr-hx: What do you think about doing these like:
```
var debug = {
  include: require("debug")("jsonApi:include")
};
```
which will make it nicely namespaced if we use more than one in the same file?

hxoliverrumbelow: Good call. I'll make those changes :+1:

pmcnr-hx:
:smile_cat: :+1: :shipit:

PR #56, opened 2015-12-01
handlerEnforcer cleanup

hxoliverrumbelow:
Fixing some tech debt by cleaning up some code duplication in the handlerEnforcer module.

### DIFF
--- a/lib/handlerEnforcer.js
+++ b/lib/handlerEnforcer.js
@@ -11,52 +11,38 @@ handlerEnforcer.wrap = function(handlers) {
   handlers.delete = handlerEnforcer._delete(handlers);
 };
 
-handlerEnforcer._search = function(handlers) {
-  var original = handlers.search;
-  return function(request, callback) {
-    original.call(handlers, request, function(err, resources) {
-      debug.handler.search(JSON.stringify(request.params), JSON.stringify(err), JSON.stringify(resources));
-      return callback(err, resources);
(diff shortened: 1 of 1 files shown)


lib/handlerEnforcer.js
```diff
@@ -11,52 +11,38 @@ handlerEnforcer.wrap = function(handlers) {
   handlers.delete = handlerEnforcer._delete(handlers);
 };
 
-handlerEnforcer._search = function(handlers) {
-  var original = handlers.search;
-  return function(request, callback) {
-    original.call(handlers, request, function(err, resources) {
-      debug.handler.search(JSON.stringify(request.params), JSON.stringify(err), JSON.stringify(resources));
-      return callback(err, resources);
+handlerEnforcer._strictReplace = function(handlers, operation, outCount) {
```
pmcnr-hx: Wonderful implementation. Always a pleasure see elegant meta-programming! :smiley_cat: The name confuses me though. I would have expected `wrapper` /  `decorator` and `factory` somewhere in the name, which would have made more obvious that this was a factory of function decorators. `_strictReplace` invokes string replacement for some reason... What do you think?

pmcnr-hx:
I like it: :+1:!

lib/handlerEnforcer.js
hxoliverrumbelow: You're right, I'll rename it :+1:

PR #65, opened 2015-12-06
Pagination!

hxoliverrumbelow:
Here's my attempt at implementing Pagination. It's using the offset+limit strategy and you can try it like this:
http://localhost:16006/rest/articles?page[offset]=0&page[limit]=1

In the resulting payload you'll see a new `page` block in the top level meta block, and paginations links with the top level `links` block:
```
{
  jsonapi: {
    version: "1.0"
  },
  meta: {
    description: "This block shows up in the root node of every payload",
    page: {
      offset: 0,
      limit: 1,
      total: 4
    }
  },
  links: {
    self: "http://localhost:16006/rest/articles?page%5Boffset%5D=0&page%5Blimit%5D=1",
    last: "http://localhost:16006/rest/articles?page%5Boffset%5D=3&page%5Blimit%5D=1",
    next: "http://localhost:16006/rest/articles?page%5Boffset%5D=1&page%5Blimit%5D=1"
  },
  data: [ ... ]
}
```

### DIFF
--- a/.tern-project
+++ b/.tern-project
@@ -1,0 +1,11 @@
+{
+  "ecmaVersion": 5,
+  "libs": [],
+  "loadEagerly": [
+    "lib/**/*.js"
+  ],
+  "plugins": {
+    "node": {},
+    "modules": {}
+  }
+}
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 [![Build Status](https://travis-ci.org/holidayextras/jsonapi-server.svg?branch=master)](https://travis-ci.org/holidayextras/jsonapi-server)
 [![npm version](https://badge.fury.io/js/jsonapi-server.svg)](http://badge.fury.io/js/jsonapi-server)
 [![Code Climate](https://codeclimate.com/github/holidayextras/jsonapi-server/badges/gpa.svg)](https://codeclimate.com/github/holidayextras/jsonapi-server)
+[![Codacy Badge](https://api.codacy.com/project/badge/grade/dd604d6548d8467caa224f078d59b182)](https://www.codacy.com/app/oliver-rumbelow/jsonapi-server)
 [![Dependencies Status](https://david-dm.org/holidayextras/jsonapi-server.svg)](https://david-dm.org/holidayextras/jsonapi-server)
 
 # jsonapi-server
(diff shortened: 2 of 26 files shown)


hxoliverrumbelow:
Addresses #8

richardschneider:
If I remember the spec correctly, you also need the links `prev`, `next`, `first` and last`.  If any are not applicable then set them to `null` or omit them.  IMHO, `first` should always be included and `last` only if the search is guaranteed to always return the same results (even when new resources are between various search fetches). 

I maybe WRONG (haven't read the code) but I think you may have confused `last` with `prev`.

hxoliverrumbelow:
The spec currently says: (http://jsonapi.org/format/#fetching-pagination)
```
Pagination

A server MAY choose to limit the number of resources returned in a response to a subset ("page") of the whole set available.

A server MAY provide links to traverse a paginated data set ("pagination links").

Pagination links MUST appear in the links object that corresponds to a collection. To paginate the primary data, supply pagination links in the top-level links object. To paginate an included collection returned in a compound document, supply pagination links in the corresponding links object.

The following keys MUST be used for pagination links:

first: the first page of data
last: the last page of data
prev: the previous page of data
next: the next page of data
Keys MUST either be omitted or have a null value to indicate that a particular link is unavailable.

Concepts of order, as expressed in the naming of pagination links, MUST remain consistent with JSON API's sorting rules.

The page query parameter is reserved for pagination. Servers and clients SHOULD use this key for pagination operations.
```

The links I've implemented in this PR appear like this:
```
page 1/3 => self, last, next
page 2/3 => self, first, last, next, prev
page 3/3 => self, first, prev
```

`next` and `prev` move forwards and backwards 1x page, `first` and `last` jump to either end of the results.

hxoliverrumbelow:
Right I've added pagination to the swagger definition and written some tests to prove it all works. I've also had to implement sorting within the MemoryHandler to ensure sort order is applied BEFORE pagination.

hxoliverrumbelow:
...and there's now a mention of pagination in the documentation.

hxoliverrumbelow:
@holidayextras/frontendguild @holidayextras/backendguild this PR will protect us from includes that attempt to grab millions of records - all search requests will default to `limit 50`, each inclusion request will have the same restriction. I'll be looking into tweaking `jsonapi-client` to automatically depaginate result sets, but the best course of action for including across a `1-many` or `many-1` relation will be to do it via code.

hxoliverrumbelow:
I'm going to merge this work and push ahead with another issue before we ship the next release.

Issue #64, opened 2015-12-06
New filtering recommendation?


hxoliverrumbelow:
I've just seen this: http://jsonapi.org/recommendations/#filtering which says, as of right now:
```
The base specification is agnostic about filtering strategies supported by a server. The filter query parameter is reserved to be used as the basis for any filtering strategy.

It's recommended that servers that wish to support filtering of a resource collection based upon associations do so by allowing query parameters that combine filter with the association name.

For example, the following is a request for all comments associated with a particular post:

GET /comments?filter[post]=1 HTTP/1.1
Multiple filter values can be combined in a comma-separated list. For example:

GET /comments?filter[post]=1,2 HTTP/1.1
Furthermore, multiple filters can be applied to a single request:

GET /comments?filter[post]=1,2&filter[author]=12 HTTP/1.1
```

In master right now, we're using one query parameter to query against a resources attributes and a different one to query against a resources relations, something like this:

```
GET /comments?relationships[author]=ad3aa89e-9c5b-4ac9-a652-6670f9f27587
```
```
GET /comments?filter[timestamp]=2017-01-02
```

Splitting the two out simplified the initial implementation, but we should move to align with the recommendation. All our `relationships` links are generated by us for `related` relationship links on foreign relations.

PR #66, opened 2015-12-08
Remove the relationships param

hxoliverrumbelow:
This addresses #64 - stop namespacing the filtering of relationships into a `relationships` query parameter and merge that functionality into the `filter` query parameter.

### DIFF
--- a/lib/postProcessing/filter.js
+++ b/lib/postProcessing/filter.js
@@ -1,10 +1,11 @@
 "use strict";
 var filter = module.exports = { };
 
+var _ = require("underscore");
 var debug = require("../debugging.js");
 
 filter.action = function(request, response, callback) {
-  var allFilters = request.params.filter;
+  var allFilters = _.extend({ }, request.params.filter);
   if (!allFilters) return callback();
 
   var filters = { };
--- a/lib/postProcessing/filter.js
+++ b/lib/postProcessing/filter.js
@@ -62,14 +62,50 @@ filter._filterMatches = function(textToMatch, propertyText) {
 filter._filterKeepObject = function(someObject, filters) {
   for (var k in filters) {
     var whitelist = filters[k].split(",");
-    var propertyText = (someObject.attributes[k] || "");
-    if (k === "id") propertyText = someObject.id;
-    var matchOR = false;
-    for (var j = 0; j < whitelist.length; j++) {
-      var textToMatch = whitelist[j];
-      if (filter._filterMatches(textToMatch, propertyText)) matchOR = true;
+
+    if (someObject.attributes.hasOwnProperty(k) || (k === "id")) {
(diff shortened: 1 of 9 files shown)


lib/postProcessing/filter.js
```diff
@@ -62,14 +62,50 @@ filter._filterMatches = function(textToMatch, propertyText) {
 filter._filterKeepObject = function(someObject, filters) {
   for (var k in filters) {
```
pmcnr-hx: Can we rename `k` to `attributeName` or something similar that makes clear what it is?

hxoliverrumbelow: Absolutely :+1:

DuncanFenning:
Looks good :shipit:

lib/postProcessing/filter.js
```diff
@@ -1,10 +1,11 @@
 "use strict";
 var filter = module.exports = { };
 
+var _ = require("underscore");
```
pmcnr-hx: Worth switching to `lodash`?

lib/postProcessing/filter.js
```diff
@@ -1,10 +1,11 @@
...
 var debug = require("../debugging.js");
 
 filter.action = function(request, response, callback) {
-  var allFilters = request.params.filter;
+  var allFilters = _.extend({ }, request.params.filter);
```
pmcnr-hx: `_.clone(request.params.filter)` should achieve the same result for this case and read better. What do you think?

lib/postProcessing/filter.js
```diff
@@ -60,16 +61,52 @@ filter._filterMatches = function(textToMatch, propertyText) {
...
+      if (!relationshipMatches) return false;
     }
-    if (!matchOR) return false;
   }
   return true;
 };
+
+filter._attributesMatchesOR = function(attributeValue, whitelist) {
+  var matchOR = false;
+  for (var j = 0; j < whitelist.length; j++) {
```
pmcnr-hx: `whitelist.forEach` instead and avoid the need for the index `j`?

lib/postProcessing/filter.js
hxoliverrumbelow: Im up for it, but the move would be better done in its own pull request - we've got several other uses of underscore dotted around at the moment.

pmcnr-hx:
This is great stuff. Just some very minor suggestions, that aren't blockers. :+1: :shipit:

lib/postProcessing/filter.js
hxoliverrumbelow: I'll change this when we move to `lodash`, lodash.clone === underscore.extend

lib/postProcessing/filter.js
hxoliverrumbelow: :+1:

PR #67, opened 2015-12-09
Recursive Includes

hxoliverrumbelow:
This partly addresses https://github.com/json-api/json-api/issues/940

I've adjusted the example resources to make `tags` recursive - tags can have parent tags, and as such, tags belong to many tags. Take a look at the changes to `example/resources/tags.js` to see the schema changes.

We can query through this recursive relationship like this:

http://localhost:16006/rest/tags/7541a4de-4986-4597-81b9-cf31b6762486?include=parent.parent.parent.parent.articles

With this work, we can now query like this:

http://localhost:16006/rest/tags/7541a4de-4986-4597-81b9-cf31b6762486?include=parent[4].articles

And we can query like this:

http://localhost:16006/rest/tags/2a3bdea4-a889-480d-b886-104498c86f69?include=[articles.tags.children][2]

which is the equivalent to:

http://localhost:16006/rest/tags/2a3bdea4-a889-480d-b886-104498c86f69?include=articles.tags.children.articles.tags.children

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 coverage.html
 complexity
 npm-debug.log
+jsonapi-server.cpuprofile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,3 +50,14 @@
 2015-11-25 - Debugging functionality
 2015-11-25 - Validation around handlers behaviour
 2015-11-25 - v0.16.0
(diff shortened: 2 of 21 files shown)


hxoliverrumbelow:
I've still got this use case to consider:
```
GET /users/301?include=[friends.user][3] HTTP/1.1
```
and after pondering this use case:

PR #68, opened 2015-12-09
Automatic CPU profiling over test suite

hxoliverrumbelow:
You'll need to be using a recent version of NodeJS for this to work!

Whenever we run the test suite, it'll automatically run a CPU profile and output it to `./jsonapi-server.cpuprofile`

If you start up `node-inspector` via `$ ./node_modules/.bin/node-inspector`, click on the "Profiles" tab, click the "Load" button and select the previously mentioned cpu profile file. You'll be able to see the profiling data in all its glory!

### DIFF
--- a/lib/postProcess.js
+++ b/lib/postProcess.js
@@ -3,7 +3,6 @@ var postProcess = module.exports = { };
 
 var jsonApi = require("..");
 var debug = require("./debugging.js");
-var _ = require("underscore");
 var externalRequest = require("request").defaults({
   pool: { maxSockets: Infinity }
 });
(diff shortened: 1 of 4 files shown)


lib/postProcess.js
```diff
@@ -87,14 +86,12 @@ postProcess.fetchForeignKeys = function(request, items, schema, callback) {
...
-      return settings && settings.__as;
-    }).forEach(function(schemaProperty) {
-      item[schemaProperty.name] = undefined;
-    });
+    for (var i in schema) {
+      var settings = schema[i]._settings;
+      if (settings && settings.__as) {
+        item[i] = undefined;
+      }
+    }
```
hxoliverrumbelow: This small changes makes the entire test suite run ~10% faster!

PR #69, opened 2015-12-10
Adding a complex recursive relationship

hxoliverrumbelow:
This adds a nice recursive relationship between all tags.

It also fixes two bugs - one around having a resource include itself via a differently named attribute, and one around attempting to include across a relationship whose value is `null`/`undefined`.

I've added some tests to prove inclusions work in the above scenarios.


hxoliverrumbelow:
This takes most of our learnings from #67 but without the expansion of recursive includes.

pmcnr-hx:
Visual inspection looks good: :+1:

DuncanFenning:
Nice :+1:

PR #73, opened 2015-12-11
Test harness improvements

hxoliverrumbelow:
This PR adds some additional checks before and after the core test suite to ensure all resources are valid both before and after the test suite has run.

It also moves all validation errors behind `DEBUG=jsonApi:validationError`.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -54,7 +54,7 @@ app.disable("etag");
 
 var requestId = 0;
 app.route("*").all(function(req, res, next) {
-  debug.requestCounter(requestId++, req.url);
+  debug.requestCounter(requestId++, req.method, req.url);
   if (requestId > 1000) requestId = 0;
   next();
 });
--- a/lib/routes/addRelation.js
+++ b/lib/routes/addRelation.js
@@ -31,7 +31,7 @@ addRelationRoute.register = function() {
         resourceConfig.handlers.find(request, callback);
       },
       function(ourResource, callback) {
-        theirResource = ourResource;
+        theirResource = JSON.parse(JSON.stringify(ourResource));
 
         var theirs = request.params.data;
         theirResource[request.params.relation] = theirResource[request.params.relation] || [ ];
--- a/package.json
+++ b/package.json
@@ -44,12 +44,12 @@
     "node-inspector": "0.12.5"
   },
   "scripts": {
-    "test": "node ./node_modules/mocha/bin/mocha -R spec ./test/*.js",
+    "test": "node ./node_modules/mocha/bin/mocha -S -R spec ./test/*.js",
     "start": "node example/server.js",
-    "coveralls": "node ./node_modules/mocha/bin/mocha --require blanket --reporter mocha-lcov-reporter ./test/*.js | node ./node_modules/coveralls/bin/coveralls.js",
-    "coverage": "node ./node_modules/mocha/bin/mocha --require blanket --reporter html-cov ./test/*.js > coverage.html",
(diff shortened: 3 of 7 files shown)


lib/router.js
```diff
@@ -54,7 +54,7 @@ app.disable("etag");
 
 var requestId = 0;
 app.route("*").all(function(req, res, next) {
-  debug.requestCounter(requestId++, req.url);
+  debug.requestCounter(requestId++, req.method, req.url);
```
hxoliverrumbelow: I've added the http verb to the request debugging, making it easier to differentiate between GET and PATCH.

lib/routes/addRelation.js
```diff
@@ -31,7 +31,7 @@ addRelationRoute.register = function() {
         resourceConfig.handlers.find(request, callback);
       },
       function(ourResource, callback) {
-        theirResource = ourResource;
+        theirResource = JSON.parse(JSON.stringify(ourResource));
```
hxoliverrumbelow: This closes a bug identified by the test harness improvements, whereby we were modifying an object provided to us by the MemoryHandler, thus mutating the object in our store prematurely.

package.json
```diff
@@ -44,12 +44,12 @@
...
-    "test": "node ./node_modules/mocha/bin/mocha -R spec ./test/*.js",
+    "test": "node ./node_modules/mocha/bin/mocha -S -R spec ./test/*.js",
     "start": "node example/server.js",
-    "coveralls": "node ./node_modules/mocha/bin/mocha --require blanket --reporter mocha-lcov-reporter ./test/*.js | node ./node_modules/coveralls/bin/coveralls.js",
-    "coverage": "node ./node_modules/mocha/bin/mocha --require blanket --reporter html-cov ./test/*.js > coverage.html",
+    "coveralls": "node ./node_modules/mocha/bin/mocha -S --require blanket --reporter mocha-lcov-reporter ./test/*.js | node ./node_modules/coveralls/bin/coveralls.js",
+    "coverage": "node ./node_modules/mocha/bin/mocha -S --require blanket --reporter html-cov ./test/*.js > coverage.html",
     "complexity": "node ./node_modules/plato/bin/plato -r -d complexity lib",
-    "performance": "node --allow-natives-syntax --harmony ./node_modules/mocha/bin/_mocha --reporter mocha-performance ./test/*.js",
+    "performance": "node --allow-natives-syntax --harmony ./node_modules/mocha/bin/_mocha -S --reporter mocha-performance ./test/*.js",
```
hxoliverrumbelow: The `-S` argument tells Mocha to sort the test files and run them in alphabetical order. This ensure the two new tests run before and after all other tests.

test/_preResourceValidationCheck.js
```diff
@@ -0,0 +1,37 @@
...
+    describe("Searching for " + resource.name, function() {
+      it("should find " + resource.count, function(done) {
+        var url = "http://localhost:16006/rest/" + resource.name;
+        helpers.request({
+          method: "GET",
+          url: url
+        }, function(err, res, json) {
+          assert.equal(err, null);
+          json = helpers.validateJson(json);
+          assert.equal(json.data.length, resource.count, "Expected " + resource.count + " resources");
```
hxoliverrumbelow: If any objects fail to validate, they'll be dropped and these assertions will fail.

DuncanFenning:
Awesome :+1:

lib/routes/addRelation.js
pmcnr-hx: Are we sure `ourResource` won't have circular references?

pmcnr-hx:
Good stuff! :+1:

lib/routes/addRelation.js
hxoliverrumbelow: That resource is the one that goes out the door to the client, so if it does have a circular include it'll crash further up the stack anyway.

PR #75, opened 2015-12-17
v1.0.3 - Fixing some regressions

hxoliverrumbelow:
After revisiting https://github.com/holidayextras/jsonapi-client, bumping the version of `jsonapi-server` and running the test suite, we identified two minor regressions:

1. `Content-Length` headers shouldn't be forwarded on for inclusion requests
2. Resources with undefined properties shouldn't match using the stock filtering functionality.

In the future I'd like to find a way of running dependent projects against new releases to identify these edge cases before each new release. I'd also like to write some extensive unit tests, but I don't think it's a good use of my time right now.

You can test this by checking out jsonapi-client, updating to the latest jsonapi-server release (v1.0.2) and running npm-test. Update to this version and the tests will pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,3 +63,6 @@
 2015-12-10 - v1.0.1
 2015-12-11 - Test harness improvements
 2015-12-11 - v1.0.2
+2015-12-17 - Don't pass on content-length HTTP headers on inclusion requests
+2015-12-17 - Filter won't match undefined properties anymore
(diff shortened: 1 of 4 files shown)


connormeredith:
Taking review

DuncanFenning:
Makes sense :+1:

connormeredith:
:+1: Tests now passing on the client when using this branch.

PR #79, opened 2015-12-30
Enable pathless api

theninj4:
This PR enables a user to drop this line from their config:
```
  base: "rest", // A path prefix
```
to serve up an API directly from the hostname.

To see this in action, delete the [base definition](https://github.com/holidayextras/jsonapi-server/blob/master/example/server.js#L26) from the example server, navigate to [http://localhost:16006/articles](http://localhost:16006/articles) and click around, verifying all the links still work.


connormeredith:
:+1:

DuncanFenning:
looks good :+1:

PR #80, opened 2015-12-31
Correctly set the Location header when creating resources

theninj4:
According to the [JSON-API spec. 201 Create](http://jsonapi.org/format/#crud-creating-responses-201)
the response header for create should be:
````
HTTP/1.1 201 Created
Location: http://example.com/photos/008ef600-0cad-4fb5-860a-958414defde7
````

Before this PR, the Location header was `http://example.com/photos`. This work appends the resource ID on the end and adds an assertion to prove it works as-per the spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,3 +68,5 @@
 2015-12-17 - v1.0.3
 2015-12-30 - Enable path-less API
 2015-12-30 - v1.0.4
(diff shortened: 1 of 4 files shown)


DuncanFenning:
:+1:

hx-damian-casey:
:+1:

Issue #84, opened 2016-01-05
Joi default() not functioning


futurist:
using mongodb-store, After set Joi default value as below:
````
......
attributes: {
    time: api.Joi.date().default(Date.now, 'time of creation'),
    age: api.Joi.number().default(18),
......
````
if omit the time/age when `POST`,  the database don't have default value specified above.

theninj4:
This is correct, we're only using Joi to pick up validation errors at the moment. It should be fairly simple to take the Joi-adjusted object after passing validation.

futurist:
I see. and please check the PR #85  above, it's worked here, but don't strictly tested.

theninj4:
Ah amazing! Thanks for the contribution, sorry I didn't see your PR sooner, I've been crazy busy over the last few days.

PR #85, opened 2016-01-06
FIX: make Joi default() function work in jsonApi.define()

futurist:
PR for Issue: [#84](https://github.com/holidayextras/jsonapi-server/issues/84)

### DIFF
--- a/example/resources/comments.js
+++ b/example/resources/comments.js
@@ -10,8 +10,8 @@ jsonApi.define({
     body: jsonApi.Joi.string()
       .description("The tag name")
       .example("Summer"),
-    timestamp: jsonApi.Joi.date().format("YYYY-MM-DD")
-      .description("The date on which the comment was created, YYYY-MM-DD")
+    timestamp: jsonApi.Joi.string()
+      .description("The date on which the comment was created, YYYY-MM-DD (TODO: change to date() format)")
       .example("2017-05-01"),
     author: jsonApi.Joi.one("people")
       .description("The person who wrote the comment"),
--- a/lib/responseHelper.js
+++ b/lib/responseHelper.js
@@ -36,13 +36,13 @@ responseHelper._enforceSchemaOnArray = function(items, schema, callback) {
 
 responseHelper._enforceSchemaOnObject = function(item, schema, callback) {
   debug.validationOutput(JSON.stringify(item));
-  Joi.validate(item, schema, function (err, value) {
+  Joi.validate(item, schema, function (err, result) {
     if (err) {
       debug.validationError(err.message, JSON.stringify(item));
       return callback(null, null);
     }
 
-    var dataItem = responseHelper._generateDataItem(value, schema);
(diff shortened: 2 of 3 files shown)


theninj4:
Something odd seems to be going on here.

On the CI server, the tests fail with different date formats:
```
-    "timestamp": "2017-06-29T00:00:00.000Z"
+    "timestamp": "2017-06-29"
```

On my local machine, the same tests fail but they're an hour different:
```
-    "timestamp": "2017-06-28T23:00:00.000Z"
+    "timestamp": "2017-06-29"
```

This timestamp field is defined in example/resources/comments.js:13
```javascript
  timestamp: jsonApi.Joi.date().format("YYYY-MM-DD")
```

After poking around a bit, I've found we're passing around a date-string formatted as "2017-06-29" - this goes through Joi and gets converted to a javascript Date, which eventually gets converted back to JSON as it leaves the API, thus appearing as a full JSON timestamp regardless of the requested format. I'm guessing the time difference I'm seeing is timezone related?

futurist:
also noticed that, firstly I think it's maybe moment.js lib issue, 
https://github.com/hapijs/joi/blob/master/lib/date.js#L74
but don't go deep at that moment.

if ISODate maybe it's will be OK?

theninj4:
For reference - I'm in the UK, so I'm currently in the GMT timezone (GMT === UTC). During the summer months of the year we move into the BST timezone (BST === UTC+1). The difference in time is caused by the input date not containing a timezone (it's "2016-06-29", not "2016-06-29T00:00:00.000Z+000").

theninj4:
Assuming the date's are correct, my next question is "why do the Date()s not conform to the requested format?".

Maybe the timestamp attribute should be re-typed to be `string`? It would at least prevent the type conversion from taking place. This might end up being a documentation note around what to expect when using the `Joi.date()` type. Or maybe we should steer clear of allowing Joi to convert types all together?

It's getting really late over here, I'm going to sleep on this one.

futurist:
so moment.js convert user input into ISO Date

http://momentjs.com/docs/#/displaying/as-javascript-date/

Joi use it as below:

 `moment('2017-06-29', 'YYYY-MM-DD', true).toDate()`

the result output is 2017-06-28T23:00:00.000Z from the CI server according the timezone.

Date & time is complicated, so maybe it's better to pass it to Joi+Moment.js to manage and convert user input into more standard format?

futurist:
I'm in China, the output is Thu Jun 29 2017 00:00:00 GMT+0800 (HKT), the same input date string, but if added to different timezone, it's different time.

theninj4:
I think you're right, what it's doing is correct, it's the Joi validation defined in the comment resource that is wrong. Could you change the validation from a date to a string? It'd be great to add a new attribute (or alter an existing attribute) to contain a default value, then tweak one of the tests to ensure the default value is being applied correctly.

futurist:
change to string() triggered many test case problems, so maybe solve the CI problems later.

PR #87, opened 2016-01-08
Use Jois sanitisation as well as validation for input+output

theninj4:
In the past we've been using Joi to simply validate objects. Joi will also sanitise data, which does some useful things like:

1. Query string properties (which are always interpreted as strings) will be coerced into the correct type - so number types in a query string will become numbers.
2. We can apply default values to both attribute and search parameters.

You can test this by starting the example server with debugging:
```
$ DEBUG=jsonApi:handler:search npm start
```
and visiting http://localhost:16006/rest/articles?query=123

You'll see in the console the type coercion taking place - the `query` param becomes a number:
```
  jsonApi:handler:search {"type":"articles","query":123,"page":{"offset":0,"limit":50}} .......
```

You'll see in the response payload every article has a new attribute `status: "published"` which is added by the new Joi schema `status: jsonApi.Joi.string().default("published")`.

### DIFF
--- a/example/resources/articles.js
+++ b/example/resources/articles.js
@@ -20,6 +20,9 @@ jsonApi.define({
     created: jsonApi.Joi.date().format("YYYY-MM-DD")
       .description("The date on which the article was created, YYYY-MM-DD")
       .example("2017-05-01"),
+    status: jsonApi.Joi.string().default("published")
+      .description("The status of the article - draft, ready, published")
+      .example("published"),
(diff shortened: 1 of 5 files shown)


theninj4:
This resolves #84 @futurist

pmcnr-hx:
Does what it says on the tin. Code looks good: :+1:

jackdcrawford:
:+1:

Issue #59, opened 2015-12-01
Build a swanky github.io website


hxoliverrumbelow:
Via github pages, build a pretty website for this project.

silentham:
Good idea dude

PR #88, opened 2016-01-09
Additional usage documentatinon

theninj4:
I've added some docs around how to migrate from an express-based API to jsonapi-server, plus a suggestion on how to organise a jsonapi-server codebase.

### DIFF
--- a/documentation/api-migration.md
+++ b/documentation/api-migration.md
@@ -24,7 +24,7 @@ var bootstrapJsonApi = function(expressApp) {
   // Alter the expressJs constructor to return your existing instance
   // preventing jsonapi-server from trying to build a new one
   require.cache[express].exports = function() {
-    return app;
+    return expressApp;
   };
   // Now load jsonapi-server, which will require express and call The
   // constructor, which now returns your existing express instance
--- a/documentation/api-migration.md
+++ b/documentation/api-migration.md
@@ -1,0 +1,38 @@
+
+### Migrating towards jsonapi-server from ExpressJS
+
+If you've got an existing API powered by ExpressJS and you want to migrate towards using json:api, you're in the right place. The migration process we're running with is to:
+
+1. Get jsonapi-server running alongside an existing express app using a path prefix.
+2. Define some new resources using the in-memory handlers.
+3. Lock in the resource definitions and develop custom handlers to provide data in the new formats by re-using existing functionality in the current codebase.
+4. Migrate traffic over to the json:api resources.
+5. Kill off the old express server.
+
+We're using something similar to the below snippet to run jsonapi-server alongside our existing express application:
+
+```javascript
+// BEFORE you run this function you should bring up your expressJs
+// server as usual, bind your routes, etc etc.
+// Do NOT require() jsonapi-server, this function does it for you!
+// Execute this function with an existing express() instance:
+var bootstrapJsonApi = function(expressApp) {
+  // Find the express module in the NodeJS internal cache
+  var express = Object.keys(require.cache).filter(function(i) {
+    return i.match('/node_modules/express/index.js');
+  }).pop();
+  // Alter the expressJs constructor to return your existing instance
+  // preventing jsonapi-server from trying to build a new one
+  require.cache[express].exports = function() {
+    return app;
+  };
+  // Now load jsonapi-server, which will require express and call The
+  // constructor, which now returns your existing express instance
+  jsonApi = require('jsonapi-server');
+  // Finally make this function call to register the json:api routes
+  require('jsonapi-server/lib/routes/index.js').register();
+
+  // Now define your resources as usual.
+  // DO NOT call jsonApi.start() !!
+};
+```
--- a/documentation/suggested-project-structure.md
+++ b/documentation/suggested-project-structure.md
@@ -64,14 +64,14 @@ jsonApi.define({
 
 #### Example handler
 
-Handlers are like your controllers - they need to, at very least, provide the functionality described in the [handler documentation](handlers.md). The follow example covers how you implement [jsonapi-store-elasticsearch](https://github.com/holidayextras/jsonapi-store-elasticsearch).
+Handlers are like your controllers - they need to, at very least, provide the functionality described in the [handler documentation](handlers.md). The following example covers how you implement [jsonapi-store-elasticsearch](https://github.com/holidayextras/jsonapi-store-elasticsearch).
 
 ```javascript
 var EsHandler = require("jsonapi-store-elasticsearch");
 
 // You might want to pull this out further to share it
 // amongst other handlers
-var specificEsCluster = var new ElasticsearchStore({
+var specificEsCluster = var new EsHandler({
   host: "localhost:9200"
 });
 
--- a/documentation/suggested-project-structure.md
+++ b/documentation/suggested-project-structure.md
@@ -45,7 +45,7 @@ jsonApi.start();
 
 #### Example resource
 
-The idea is to stick to having one resource per file and stick to pure config. Each file should do nothing more than define a resource. Handlers should be referenced from the `../handlers` folder, which enables us to easily abstract functionality by sharing features amongst handlers. Resource files are effectively defining your routing  layer.
+The idea is to stick to having one resource per file and stick to pure config. Each file should do nothing more than define a resource. Handlers should be referenced from the `../handlers` folder, which enables us to easily abstract functionality by sharing features amongst handlers. Resource files are effectively defining your routing layer.
 
 ```javascript
 var commentHandler = require("../handlers/commentHandler.js");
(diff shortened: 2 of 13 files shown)


documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
...
+
+// You might want to pull this out further to share it
+// amongst other handlers
+var specificEsCluster = var new ElasticsearchStore({
+  host: "localhost:9200"
+});
+
+module.exports = specificEsCluster;
+
+// // If you want to intercept calls to extend existing behaviour:
```
connormeredith: What's the deal with this?

connormeredith:
:+1: looks good dude

documentation/api-migration.md
```diff
@@ -0,0 +1,38 @@
...
+// Execute this function with an existing express() instance:
+var bootstrapJsonApi = function(expressApp) {
+  // Find the express module in the NodeJS internal cache
+  var express = Object.keys(require.cache).filter(function(i) {
+    return i.match('/node_modules/express/index.js');
+  }).pop();
+  // Alter the expressJs constructor to return your existing instance
+  // preventing jsonapi-server from trying to build a new one
+  require.cache[express].exports = function() {
+    return app;
```
lukehansell: did you mean `expressApp`

documentation/api-migration.md
```diff
@@ -0,0 +1,38 @@
...
+  };
+  // Now load jsonapi-server, which will require express and call The
+  // constructor, which now returns your existing express instance
+  jsonApi = require('jsonapi-server');
+  // Finally make this function call to register the json:api routes
+  require('jsonapi-server/lib/routes/index.js').register();
```
lukehansell: could you not just use `jsonapi-server/lib/routes`?

documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
+### Suggested Project Structure
+
+ * `server.js` is the main entry point.
+ * Think of the `resources` folder as your `routes`.
```
lukehansell: If these are synonymous then why aren't they named as such? I think these two lines might be superfluous. Could you instead describe what they do and allow the reader to make the connection?

documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
...
+
+This is the main entry point for your API. The goal is to configure jsonapi-server and load all the resources.
+
+```javascript
+var jsonApi = require("jsonapi-server");
+var fs = require("fs");
+var path = require("path");
+
+jsonApi.setConfig({
+  // Put your config here!
```
lukehansell: do you have examples of what the config might look like? e.g what params are required? You could write:
```
jsonApi.setConfig({
  // host: 'http://foobar.com,
  // port: '5000',
  // foo: 'bar'
})
```

By providing the required config with example values, but commented out, people can more easily pick this up and run with it.

documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
...
+});
+
+// Load all the resources
+fs.readdirSync(path.join(__dirname, "/resources")).filter(function(filename) {
```
lukehansell: this overcomplicates what is happening here. Since this is a tutorial could we do this the simplest way possible and allow people to add these niceties afterwards? I always find the more direct and simple a tutorial the easier it is to follow.

documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
...
+    url: jsonApi.Joi.string().uri()
+    height: jsonApi.Joi.number().min(1).max(10000).precision(0)
+    width: jsonApi.Joi.number().min(1).max(10000).precision(0)
+  }
+});
+```
+
+#### Example handler
+
+Handlers are like your controllers - they need to, at very least, provide the functionality described in the [handler documentation](handlers.md). The follow example covers how you implement [jsonapi-store-elasticsearch](https://github.com/holidayextras/jsonapi-store-elasticsearch).
```
lukehansell: "The follow example" => "The following example"

documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
...
+
+```javascript
+var EsHandler = require("jsonapi-store-elasticsearch");
+
+// You might want to pull this out further to share it
```
lukehansell: could you describe what and why you're doing here?

documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
...
+// amongst other handlers
+var specificEsCluster = var new ElasticsearchStore({
```
lukehansell: unnecessary `var`?
and should `ElasticsearchStore` be `EsHandler`?

documentation/suggested-project-structure.md
```diff
@@ -0,0 +1,97 @@
...
+}).map(function(filename) {
+  return path.join(__dirname, "/resources/", filename);
+}).forEach(require);
+
+jsonApi.start();
+```
+
+#### Example resource
+
+The idea is to stick to having one resource per file and stick to pure config. Each file should do nothing more than define a resource. Handlers should be referenced from the `../handlers` folder, which enables us to easily abstract functionality by sharing features amongst handlers. Resource files are effectively defining your routing layer.
```
lukehansell: I was wondering if the following might be clearer?
```
Resource files define the configuration of an item including what it looks like. We define a [handlers](link to handler section) which denotes where the data is stored. By referencing handlers externally from `../handlers` we can share these among many resources.
```

documentation/api-migration.md
theninj4: Yep, good spot :+1:

documentation/api-migration.md
theninj4: I'd rather be explicit about the file, I've got a pet hate for targeting a folder that happens to have a `package.json` that sends me somewhere else. By looking at this line of code and seeing `index.js` it makes debugging this delightful hack that little bit easier.

documentation/suggested-project-structure.md
theninj4: Yep it's here: https://github.com/holidayextras/jsonapi-server/blob/master/documentation/configuring.md

documentation/suggested-project-structure.md
theninj4: If someones going to follow this folder+file structure they'll immediately need to have something that loads all files in a given folder. Basically nothing will work without these 5x lines, unless someone has lots and lots of `require(---);` statements one after the other.

documentation/suggested-project-structure.md
theninj4: Good spot :+1:

documentation/suggested-project-structure.md
theninj4: These handlers can be written any which way you like, the functional requirements are here https://github.com/holidayextras/jsonapi-server/blob/master/documentation/handlers.md this is mainly an example I can link people to when I get issues opened around how to override/extend the ORM handlers we've written.

documentation/suggested-project-structure.md
theninj4: Gah, typo's! :+1:

documentation/suggested-project-structure.md
theninj4: Coming from the routing+controllers ideology, I think this reduces the initial barrier to entry. JSON:API enforces all routes, the idea being you only have a say in what the resources look like, so calling the folder `resources` is the correct name and saying `think of it as a routes folder` is the best way to get people onboard quickly. A similar thing applies to `handlers` - they're something I've introduced that require a special name because all handlers must conform with a spec, you can't simply expose funky custom functionality and expect it to work.

documentation/suggested-project-structure.md
theninj4: I don't want to put too much more work time into this, I'd rather merge what I've got and iterate it again in the future. Unless you think it's misleading? Remember people will start at the documentation index and possibly click through to here.

Long term I'd like to look into this #59

documentation/suggested-project-structure.md
lukehansell: maybe link to this here then?

documentation/suggested-project-structure.md
lukehansell: I understand the reasoning behind this, but those 5 require statements would read nicer and be easier to understand. Since this is only a demo and not a production piece of code I still feel this is overkill, but I won't block this PR on it.

documentation/suggested-project-structure.md
lukehansell: it's not misleading (except perhaps the last part about routing), more convoluted. I suppose "what" a resource is is defined elsewhere though?

theninj4: Yep: https://github.com/holidayextras/jsonapi-server/blob/master/documentation/resources.md

PR #89, opened 2016-01-09
Cleanup package.json

theninj4:
* Update package description
* Move `v8-profiler` from `dependencies` to `devDependencies` (it's only used by the test suite)

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,18 +19,18 @@
     "node": "*"
   },
   "dependencies": {
-    "async": "1.4.2",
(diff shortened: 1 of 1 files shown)


lukehansell:
looks good to me :+1:

theninj4:
I've updated the dependencies - if I bump Joi to v7 we lose support for old versions of node, judging by the market share of node v0.10.x I'm leaving us on the lastest v6 release.

connormeredith:
:+1:

Issue #94, opened 2016-01-18
server responds to OPTIONS method with 404


moeabm:
The browser is sending a preflight OPTIONS request that returns 404. This seems to be a CORS. I could very well be doing something wrong but cannot find any docs on how to implement CORS on the server.

moeabm:
This may be out of the scope of this project. I'm just using a reverse proxy to put the server on same host, port and protocol as the client (ember-data).

theninj4:
Ah. The current behaviour is to set the following headers on every response:
```
    "Access-Control-Allow-Origin": "*",
    "Access-Control-Allow-Methods": "*"
```
...however we're fronting all our environments with nginx which will be handling OPTIONS requests for us. jsonapi-server should be responding to the OPTIONS request.

theninj4:
The default headers should now enable CORS out-of-the-box :+1:

moeabm:
Very nice. Thank you.

PR #95, opened 2016-01-19
Respond with 204 OK + CORS to OPTIONS requests

theninj4:
I've refactored response headers a little to make it more consistent, and we're now responding with HTTP 204 instead of 404 for OPTIONS request. There's a new test to prevent any regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,3 +75,5 @@
 2015-01-11 - Updating dependencies
 2015-01-11 - Error hard and fast when incorrectly defining foreign relations
 2015-01-11 - v1.1.0
+2015-01-21 - Set better default CORS headers
+2015-01-21 - v1.2.0
--- a/lib/router.js
+++ b/lib/router.js
@@ -50,8 +50,8 @@ app.use(function(req, res, next) {
   res.set({
     "Content-Type": "application/vnd.api+json",
     "Access-Control-Allow-Origin": "*",
-    "Access-Control-Allow-Methods": "*",
-    "Access-Control-Allow-Headers": "*"
+    "Access-Control-Allow-Methods": "GET, POST, PATCH, DELETE, OPTIONS",
+    "Access-Control-Allow-Headers": req.headers["access-control-request-headers"] || ""
   });
 
   if (req.method === "OPTIONS") {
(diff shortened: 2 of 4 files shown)


theninj4:
Resolves #94

pmcnr-hx:
Code changes look good and tests pass: :+1:

fabiobon:
I've tried to merge this pull but in browser there's one error. 
XMLHttpRequest cannot load http://localhost:16006/rest/articles. Request header field Content-Type is not allowed by Access-Control-Allow-Headers.
I've tried to put 
var client = new JsonapiClient("http://localhost:16006/rest", {  
	header: {
	    "Content-Type": "application/vnd.api+json",
	    "Accept": "application/vnd.api+json"
  	}
});
with no success.
Have you got any suggestion?
Thanx

theninj4:
Thanks for taking a look @fabiobon - I was missing a CORS header, they should all be there now.

theninj4:
I've tagged this as work-in-progress pending a client-side test suite for [jsonapi-client](https://github.com/holidayextras/jsonapi-client) to prove this PR solves our CORS issues.

fabiobon:
Ok, after some tests configuration below works but, in browser, does not returns payload.body

  res.set({
    "Content-Type": "application/vnd.api+json",
    "Access-Control-Allow-Origin": "*",
    "Access-Control-Allow-Methods": "*",
    "Access-Control-Allow-Headers": "Origin, X-Requested-With, Content-Type, Accept"
  });

  if (req.method === "OPTIONS") {
    return res.status(204).end();
  }

NOTE: to run in browser without cors erros I've replaced 
"Access-Control-Allow-Headers": "*"
with
"Access-Control-Allow-Headers": "Origin, X-Requested-With, Content-Type, Accept"

fabiobon:
Browser Only:
59. return callback(null, payload.body, payload.body.data, payload.body.included);
inside Transport._action retrives ERROR:
Uncaught TypeError: Cannot read property 'data' of null

theninj4:
Thanks @fabiobon - I've fixed up this branch, it turns out we're not allowed to wildcard the `Access-Control-Allow-Headers` field, so I've resorted to simply reflecting what the client suggests. I've also had to lock down the `Access-Control-Allow-Methods`. This now resolves CORS issues, I've got a clientside test suite running for jsonapi-client in Firefox and Chrome against this branch :+1:

PR #96, opened 2016-01-25
Aggressive cache-expiry headers by default

theninj4:
I've seen some browsers (IE9 + IE11) aggressively cache API responses, so much so that this chain of requests won't behave as expected:
 1. Find resource 1234
 2. Delete resource 1234
 3. Find resource 1234 <--- still works, returns { }

This PR adds some headers so the default behaviour says "please don't ever cache this".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,3 +77,5 @@
 2015-01-11 - v1.1.0
 2015-01-21 - Set better default CORS headers
 2015-01-21 - v1.2.0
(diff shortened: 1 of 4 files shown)


theninj4:
This should resolve some of the browser testing issues over at https://github.com/holidayextras/jsonapi-client/pull/11

jackdcrawford:
Seen this working :+1:

connormeredith:
Looks good to me :+1:

PR #100, opened 2016-02-16
Filter by non-string attributes

pmcnr-hx:
Add support for correctly executing filtering operations when attributes that are not strings are involved. We now use Joi to validate the filter parameters and convert to the expected attribute type, which _should_ work for the supported Joi attribute types and make the comparisons work.

This fixes inequality filtering by attribute types such as integers and strings.

How can you test this? The new system tests cover inequality filtering by date attributes. Feel free to experiment with the sample resources definitions and test additional filtering cases.

* [x] :+1: 
* [x] :+1: :+1:

### DIFF
--- a/lib/postProcessing/filter.js
+++ b/lib/postProcessing/filter.js
@@ -2,7 +2,8 @@
 var filter = module.exports = { };
 
 var _ = {
-  assign: require("lodash.assign")
+  assign: require("lodash.assign"),
+  isEqual: require("lodash.isequal")
 };
 var debug = require("../debugging.js");
 
@@ -64,7 +65,7 @@ filter._filterMatches = function(filterElementStr, attributeValue, attributeConf
   }
   filterElement.value = validationResult.value;
   if (!filterElement.operator) {
-    return attributeValue === filterElement.value;
+    return _.isEqual(attributeValue, filterElement.value);
   }
   if (["~", ":"].indexOf(filterElement.operator) !== -1 && typeof filterElement.value !== "string") {
     return false;
--- a/lib/postProcessing/filter.js
+++ b/lib/postProcessing/filter.js
@@ -6,6 +6,8 @@ var _ = {
 };
 var debug = require("../debugging.js");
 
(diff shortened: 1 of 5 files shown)


jackdcrawford:
:mag: Reviewing

jameshhx:
Taking :one:

lib/postProcessing/filter.js
```diff
@@ -46,30 +48,54 @@ filter.action = function(request, response, callback) {
...
+filter._filterMatches = function(filterElementStr, attributeValue, attributeConfig) {
+  var filterElement = filter._splitFilterElement(filterElementStr);
+  var validationResult = attributeConfig.validate(filterElement.value);
+  if (validationResult.error) {
+    debug.filter("invalid filter condition value:", validationResult.error);
+    return false;
+  }
+  filterElement.value = validationResult.value;
+  if (!filterElement.operator) {
+    return attributeValue === filterElement.value;
```
jameshhx: The direct comparison for non-string types (such as dates) does not appear to be working.

I'd expect the following URL to return a single record titled 'How to AWS'

http://localhost:16006/rest/articles?filter[created]=2016-02-08

pmcnr-hx:
Should be fixed for date equality now @jameshhx. You'll have to strictly use the `YYYY-MM-DD` format though, because we're validating that particular format in the Joi attribute definition. I've added two additional system tests to cover the equality filtering cases.

pmcnr-hx:
After some research, dates are a complicate problem to solve, and we're not going to support them at this point in time (we're simply using ISO string representations for now). I am going to change the inequality tests to use a number attribute instead.

jameshhx:
I can see the filtering being applied correctly for dates as strings. The following URL returns a single article, with the correct `created` date.

http://localhost:16006/rest/articles?filter[created]=2016-01-05

jameshhx:
Happy for this to be merged as-is. The filtering is being applied correctly to non-string types (barring Dates for the time being). :+1:

jackdcrawford:
This looks good @pmcnr-hx :+1:

PR #101, opened 2016-02-24
Filtering falsy boolean values

DuncanFenning:
After adding a use case for booleans, falsy boolean values couldn't be filtered against.

New use cases have been added along with relevant unit tests.

This now allows filtering such as - `http://localhost:16006/rest/photos?filter[raw]=true` and `http://localhost:16006/rest/photos?filter[raw]=false`

- [x] :shipit: 
- [x] :+1:


theninj4:
Ahhh nice one! Such a subtle mistake!

jackdcrawford:
:+1:

Issue #103, opened 2016-03-01
Return all validation errors instead of only the first failure


dalane:
It would be useful if all validation errors were returned instead of only the first failure. Looking through the code it looks like this would be pretty straightforward by passing the ```{abortEarly: false}``` option to ```Joi.validate()``` and then iterating through each of the errors to create each error object.

theninj4:
Good suggestions, thanks!

PR #108, opened 2016-04-08
Return all validation errors

theninj4:
In master if we attempt to perform an action that fails validation, we will only be told about the first validation error. This PR enables us to respond with all validation errors. Here's a sample response taken from the test suite:

```javascript
{
  "jsonapi": {
    "version": "1.0"
  },
  "meta": {
    "description": "This block shows up in the root node of every payload"
  },
  "links": {
    "self": "http://localhost:16006/rest/articles"
  },
  "errors": [
    {
      "status": "403",
      "code": "EFORBIDDEN",
      "title": "Param validation failed",
      "detail": [
        {
          "message": "\"title\" is required",
          "path": "title",
          "type": "any.required",
          "context": {
            "key": "title"
          }
        },
        {
          "message": "\"content\" is required",
          "path": "content",
          "type": "any.required",
          "context": {
            "key": "content"
          }
        }
      ]
    }
  ]
}
```

### DIFF
--- a/lib/routes/helper.js
+++ b/lib/routes/helper.js
@@ -12,13 +12,13 @@ var _ = {
 
 helper.validate = function(someObject, someDefinition, callback) {
   debug.validationInput(JSON.stringify(someObject));
-  Joi.validate(someObject, someDefinition, function (err, sanitisedObject) {
+  Joi.validate(someObject, someDefinition, { abortEarly: false }, function (err, sanitisedObject) {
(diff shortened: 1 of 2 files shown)


theninj4:
This addresses #103

pmcnr-hx:
Look legit: :+1:

DuncanFenning:
This will most definitely help 👍

Issue #110, opened 2016-04-10
Question: best way to implement verbose errors on custom authenticate handler


paul42:
So, I was wondering how I can create errors in the same manner as the built-in way - 
for example, when I try to POST to a route and don't pass validation, i get a response like this:

    {
      "jsonapi": {
        "version": "1.0"
      },
      "meta": {
        "copyright": "Blah"
      },
      "links": {
        "self": "http://localhost:16006/rest/comments/"
      },
      "errors": [
        {
          "status": "403",
          "code": "EFORBIDDEN",
          "title": "Param validation failed",
          "detail": {
            "name": "ValidationError",
            "details": [
              {
                "message": "\"title\" must be a string",
                "path": "title",
                "type": "string.base",
                "context": {
                  "value": 2,
                  "key": "title"
                }
              }
            ],
            "_object": {
              "id": "cfdccd45-6395-4b44-b540-41b45b4f3746",
              "type": "comments",
              "title": 2,
              "body": "whatever!!!?"
            }
          }
        }
      ]
    }

but when I try to make my custom authenticate route respond like thus:
        var currentClientSecret = process.env.CB_CLIENT_SECRET;
    jsonApi.authenticate(function(request, callback) {
      // console.log(request.headers.authorization);
      var authorization = _.get(request, 'headers.authorization', '');
      var isBearerToken = authorization.indexOf('Bearer ') > -1;
      if(isBearerToken){
        var token = authorization.slice(7);
        console.log(token);
    
        try {
          var decoded = jwt.verify(token, new Buffer(currentClientSecret, 'base64'));
          console.log(decoded);
          callback();
        } catch (e) {
          console.log(e);
          callback(e);
        }
      } else {
        var badFormat = new Error('')
        var otherData = {
          status: 401,
          code: 'ENOAUTH',
          title:'Authorization header not in correct format. should be \'Bearer {{token}}\''
        }
        _.assign(badFormat, otherData);
        callback(badFormat);
      }
    
    });

so when I try to test it by submitting a request without the authorization header, I don't get back a verbose response, I get back just a blank 401, no body. is there a way to tap into the nice pattern from above? or should I be implementing authentication differently?

paul42:
looks like I just should just use the server and client - sorry for the spam!

PR #111, opened 2016-04-11
first pass at descriptive authenticate error?

paul42:
Not certain if this is the best way to handle it, but a first pass to start discussion. regarding issue #110 

thoughts?


paul42:
looks like I just should just use the server and client - sorry for the spam!

PR #118, opened 2016-04-28
Allow overriding debug function.

pmcnr-hx:
Add a function to the `debugging` module that allows overriding the debugging functions for each of the namespaces with a user provided function.

* [x] :heart: 
* [x] :two_hearts:
* [x] :heart: x :three:

### DIFF
--- a/documentation/debugging.md
+++ b/documentation/debugging.md
@@ -25,3 +25,21 @@ $ DEBUG=jsonApi:handler:find npm test
 ```
 $ DEBUG=jsonApi:handler:* npm test
 ```
+
+### Integration with application logging
+
+If you wish to integrate `jsonapi-server` debug output with your application's logging solution, you can override the functions that are invoked for outputting the messages for the supported namespaces by invoking the package's `debugging.__overrideDebugOutput(outputFnFactory)` function where the `outputFnFactory` parameter is a function that expects a namespace string parameter and returns an output function for the namespace.
+
+A simple example of this functionality in action would be:
+
+```javascript
+var debugging = require("jsonapi-server/lib/debugging");
+
+var outputFnfactory = function(namespace) {
+  return function(message) {
+    console.log(namespace + ">>>", message);
+  }
+}
+
+debugging.__overrideDebugOutput(outputFnfactory);
+```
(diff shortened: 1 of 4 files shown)


documentation/debugging.md
```diff
@@ -25,3 +25,7 @@ $ DEBUG=jsonApi:handler:find npm test
 ```
 $ DEBUG=jsonApi:handler:* npm test
 ```
+
+### Integration with application logging
+
+If you wish to integrate `jsonapi-server` debug output with your application's logging solution, you can override the functions that are invoked for generating the output of each of the supported namespaces by invoking the package's `debugging.__setLoggers(loggerFactory)` function where the `loggerFactory` parameter is a function that expects a namespace string parameter and returns a logger function for the namespace.
```
theninj4: Could you put an example beneath this? It's not entirely obvious by reading this paragraph and the code is both elegant and not immediately obvious.

In other words - people will ask us a lot of questions about how this works, so lets just add an example we can refer them back to.

lib/debugging.js
```diff
@@ -1,18 +1,33 @@
...
-  errors: require("debug")("jsonApi:errors"),
-  requestCounter: require("debug")("jsonApi:requestCounter")
+  include: debug("jsonApi:include"),
+  filter: debug("jsonApi:filter"),
+  validationInput: debug("jsonApi:validation:input"),
+  validationOutput: debug("jsonApi:validation:output"),
+  validationError: debug("jsonApi:validation:error"),
+  errors: debug("jsonApi:errors"),
+  requestCounter: debug("jsonApi:requestCounter"),
+  __setLoggers: setLoggersHelper.bind(null, debugging)
```
theninj4: Is `setLoggers` a good choice of function name? Should it be more like `.__redirectOutputTo(winston)` or similar?

pmcnr-hx: Agreed. I was struggling to come up with a decent name. I'll rename accordingly.

theninj4:
Works a treat :+1:

reecemillsom:
taking 3️⃣

DuncanFenning:
got :two_hearts:

reecemillsom:
👍

DuncanFenning:
go for it 👍

PR #120, opened 2016-05-04
Updated example doc for extending existing create.

dlras2:
Example documentation for intercepting calls to extend existing behavior was using an incorrect signature, causing confusion when attempting to follow the code.


theninj4:
Thanks for the help :+1:

PR #48, opened 2015-11-24
Handler enforcer

hxoliverrumbelow:
This PR enforces the expected parameters and, more importantly, the resulting values provided by custom handler implementations.

@pmcnr-hx this should make it easier to implement your mongoDb handler!

This also includes some debugging via the [debug](https://www.npmjs.com/package/debug) module, with the namespacing follows their recommendation:
```
$ DEBUG=jsonApi:handler:find npm test
```
```
$ DEBUG=jsonApi:handler:* npm test
```

PR #124, opened 2016-05-14
Prevent crash when requesting unavailable functionality

### LINKED ISSUE #106: Disabling PATCH, DELETE and POST
How to allow only the GET and OPTIONS method on a type?

theninj4:
This fixes #106 - Back in #48 we broke some functionality that would return sensible errors when an API consumer requests an unavailable CRUD feature.

With this change, if you remove the `MemoryStore.prototype.search` function within `lib/memoryHandler.js` and visit http://localhost:16006/rest/articles you'll see this:
```
{
  jsonapi: {
    version: "1.0"
  },
  meta: {
    description: "This block shows up in the root node of every payload"
  },
  links: {
    self: "http://localhost:16006/rest/articles"
  },
  errors: [
    {
      status: "403",
      code: "EFORBIDDEN",
      title: "Resource not supported",
      detail: "The requested resource 'articles' does not support 'search'"
    }
  ]
}
```

Which was the original expected behaviour.


pmcnr-hx:
Can we add a regression test for this where we disable one of the functions and verify we get the `403` error?

theninj4:
Yep, absolutely :+1:

theninj4:
@pmcnr-hx - Test added :+1:

PR #125, opened 2016-05-14
Respond with friendlier errors to unauthorised requests

### LINKED ISSUE #113: Return error details
On a 401, we are not returning any error details passed by the auth hook, just a 401 status.

theninj4:
This resolves #113 - when a client makes a request that fails the authentication check, they'll now get a friendlier response:

```
{
  jsonapi: {
    version: "1.0"
  },
  meta: {
    description: "This block shows up in the root node of every payload"
  },
  links: {
    self: "http://localhost:16006/rest/articles"
  },
  errors: [
    {
      status: "401",
      code: "UNAUTHORIZED",
      title: "Authentication Failed",
      detail: "Nope"
    }
  ]
}
```

The contents of the `detail` property come directly from the result of the authentication function, allowing developers to inject more detail about why authentication failed.


pmcnr-hx:
:+1: :shipit:

PR #126, opened 2016-05-14
Rerouter - Bypass Express on inclusions

### LINKED ISSUE #119: Included relationships fail when host and protocol are unspecified.
To reproduce:

1. Run the example server
2. Query `http://localhost:16006/articles?include=author`
3. Observe the `included` people are included
4. Remove `protocol` and `hostname` from `setConfig` in `index.js`
5. Rerun the above query
6. Observe the `included` people are now missing

While including the protocol and hostname fixes the inclusions, it also prepends it to all the links. This is fine as long as the outgoing links match the consuming service, but doing any sort of port mapping, load balancing, hostname redirects, etc, means that the exposed links do not include the expected hostnames. Leaving the links host-less allows this, and is allowed by the spec ([e.g.](http://jsonapi.org/format/#fetching-relationships))

theninj4:
This resolves #119 - All inclusions now use a `rerouter` module to bypass all HTTP routing and jump straight back in on the same infrastructure / NodeJS instance that the request began on.

This also means that inclusion's now bypass subsequent `authentication` passes, which I... *think*... is a plus. There'll also be more accurate request logs - all requests will be legitimate customer requests. We'll also be limiting the damage of any requests that might somehow cause us to DOS ourselves via inclusions.

The test suite runs ~10% faster, which is also a nice-to-have. Not that it's slow or anything, but it should yield performance benefits in production.

 - [x] :+1: Pedro
 - [x] :+1: Duncan
 - [x] :+1: Connor


lib/rerouter.js
```diff
@@ -0,0 +1,57 @@
...
+  if (path.match(/^https?\:\/\//)) {
+    path = path.split("/").slice(3).join("/");
+  }
+  if (jsonApi._apiConfig.base !== "/") {
+    if (path[0] !== "/") path = "/" + path;
+    path = path.split(jsonApi._apiConfig.base);
+    path.shift();
+    path = path.join(jsonApi._apiConfig.base);
+  }
+  path = path.replace(/^\//, "").split("?")[0].replace(/\/$/, "");
```
theninj4: All this string manipulation is unfortunate, but hard to avoid. The internal strings were designed to make it really easy to glue them together to generate user-facing URLs, not destructure user-facing URLs. Oh well.

lib/rerouter.js
```diff
@@ -0,0 +1,57 @@
...
+
+  var route = Object.keys(validRoutes).filter(function(someRoute) {
+    someRoute = someRoute.replace(/(\:[a-z]+)/g, "[^/]*?");
+    someRoute = new RegExp("^" + someRoute);
+    return someRoute.test(path);
+  }).pop();
```
theninj4: This is basically doing the ExpressJS routing layer - picking the appropriate route out of all those registered.

lib/rerouter.js
```diff
@@ -0,0 +1,57 @@
...
+    url: newRequest.uri,
+    headers: newRequest.headers,
+    cookies: newRequest.cookies,
+    params: url.parse(newRequest.uri.split("?").pop())
+  };
+
+  route.split("/").forEach(function(urlPart, i) {
+    if (urlPart[0] !== ":") return;
+    req.params[urlPart.substring(1)] = path.split("/")[i];
+  });
```
theninj4: This extracts named params from the url - `:type/:id` and `people/123` into `{ type: "people", id: "123" }`.

pmcnr-hx:
I just skimmed the code, but it's looking good. I'll do a proper review tomorrow. I also suspect this will sort some of the issues we're having in a project that is using this module.

theninj4:
:+1: I'd like you, @connormeredith and @DuncanFenning to give me the thumbs up here before we merge it.

DuncanFenning:
on it

lib/postProcessing/include.js
```diff
@@ -186,28 +184,15 @@ includePP._fillIncludeTree = function(includeTree, request, callback) {
...
-      if (!json) return done(null, [ ]);
-
-      if (res.statusCode >= 400) {
-        debug.include("!!", JSON.stringify(json));
-        return done(json.errors);
+      headers: request.safeHeaders,
+      cookies: request.cookies
+    }, function(err, json) {
+      if (err) {
+        debug.include("!!", JSON.stringify(err));
```
pmcnr-hx: Worth having a namespaced debug for `rerouter`? The above might be difficult to identify when looking at debug output otherwise.

lib/rerouter.js
pmcnr-hx: I would move all the path processing into a separate helper function.

lib/rerouter.js
pmcnr-hx: Ditto with the route processing. It's complex enough to warrant its own helper function.

lib/router.js
```diff
@@ -88,15 +88,22 @@ router.close = function() {
   server = null;
 };
 
+router._routes = { };
```
pmcnr-hx: `Object.create(null)` instead? [Objects without prototypes are better maps](http://www.2ality.com/2013/10/dict-pattern.html)

theninj4:
I've just added a new commit to ensure additional properties added within the authentication function are passed through on inclusions.

lib/router.js
theninj4: I like the theory behind what they're saying, but I disagree with us changing things:

1. The edge cases they're trying to avoid aren't applicable to us, the only place we touch prototypes in this project is the in-memory handler. 
2. I like to keep things simple.
3. I like to keep things consistent. We don't do this anywhere else in this codebase. Nor do we need to.

connormeredith:
Can confirm that this fixes the issue 👍  I agree with @pmcnr-hx, the path processing code would be much better in it's own function

DuncanFenning:
Agree with comments, 👍

theninj4:
I've pulled it out into private functions :+1:

lib/router.js
pmcnr-hx: I defer to the consistency argument. In an ES6 world where there is a proper `Map` I would push for using that though. There's no reason for abusing objects in ES6 any more.

lib/rerouter.js
theninj4: This has moved, but GitHub's comment hasn't collapsed for some strange reason :(

pmcnr-hx:
:+1: :shipit:

PR #128, opened 2016-05-17
Allow SSL/TLS connections

BryceCicada:
It's currently possible to set the protocol in config, like:

    jsonApi.setConfig({
      protocol: "https"
    });

This builds the urls in the api responses with 'https' as the url scheme.  It does not host the server with that protocol.

In my project, I want to use jsonapi-server over https.  I propose this change to allow the creation of an https server with provided tls config.  See:

- https://nodejs.org/api/https.html#https_https_createserver_options_requestlistener
- https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener

### DIFF
--- a/lib/jsonApi.js
+++ b/lib/jsonApi.js
@@ -107,7 +107,14 @@ jsonApi.onUncaughtException = function(errHandler) {
 
 jsonApi.start = function() {
   routes.register();
-  router.listen(jsonApi._apiConfig.port);
+  var serverFactory = function(app) {
+      if (jsonApi._apiConfig.protocol === 'https') {
+        return require('https').createServer(jsonApi._apiConfig.tls || {}, app);
+      } else {
+        return require('http').createServer(app);
+      }
+  };
+  router.listen(jsonApi._apiConfig.port, serverFactory);
 };
 
 jsonApi.close = function() {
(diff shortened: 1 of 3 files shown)


theninj4:
I love it. It's a great contribution, thanks!

lib/jsonApi.js
```diff
@@ -107,7 +107,14 @@ jsonApi.onUncaughtException = function(errHandler) {
 
 jsonApi.start = function() {
   routes.register();
-  router.listen(jsonApi._apiConfig.port);
+  var serverFactory = function(app) {
+      if (jsonApi._apiConfig.protocol === "https") {
+        return require("https").createServer(jsonApi._apiConfig.tls || {}, app);
+      } else {
+        return require("http").createServer(app);
```
theninj4: I find myself wondering if this could be easier to follow if we moved this `serverFactory` into `router.js` to keep all of the HTTP(s) routing code in one place - everything you need is in scope, this can simply be lifted from here into the `router.listen` function. I'm happy to wait and let you make the change, or I'm also happy to merge this as-is and I'll come back to it later?

BryceCicada: Agreed.  I'll move it.

theninj4:
Amazing, thanks @BryceCicada :+1:

Issue #129, opened 2016-05-18
swagger.json generated by the example has errors


jbockerstette:
Swagger generation doesn't seem to work correctly. The generated swagger.json contains may errors from the example project.

In order to debug this problem, I eliminated all the resources except for the people.js resource. In addition, I got rid of references to the other resources in the people.js file.

I then used the swagger editor to fix the problems found in the resulting swagger.json

I am providing the links to my people.js file : https://drive.google.com/file/d/0BzXg1kzQ457OeHpUMVhtVDk2MEk/view?usp=drivesdk

Here is the swagger.json file that was generated: https://drive.google.com/file/d/0BzXg1kzQ457ONDlVZUZZOElZdUE/view?usp=drivesdk

Here is the swagger.json file that I fixed using the Swagger editor: https://drive.google.com/file/d/0BzXg1kzQ457ORVdpdGlQRU90bVk/view?usp=drivesdk

theninj4:
Thanks for reporting this, I'll take a look and sort it out :+1:

jbockerstette:
Here is one fix, but there are more:
lib/swagger/path.js line 119
          required: [ "jsonapi", "meta, links" ],   
Should be
          required: [ "jsonapi", "meta", "links" ],

theninj4:
I'm going to investigate adding another step to the CI to validate the swagger output on each build, just in case we introduce any regressions in the future.

jbockerstette:
If you do a diff on these two files, you will see where all the changes that I made to fix my swagger.json.
In my swagger.json, there are 4 error types and 1 warning which are repeated over and over. So even though I had to make a lot of changes, there are really only 4 bugs and 1 warning.

I am adding these files
[swagger_json_broken.txt](https://github.com/holidayextras/jsonapi-server/files/275181/swagger_json_broken.txt)
[swagger_json_fixed.txt](https://github.com/holidayextras/jsonapi-server/files/275182/swagger_json_fixed.txt)

 because there is additional errors that you would not see in my first report.

theninj4:
Thanks @jbockerstette - I think I've fixed all of the bugs, and cleaned up the `required` attributes. There's also a CI build task that will validate the generated swagger docs, which should help prevent this from happening again.

jbockerstette:
You're awesome. Great job. Is there some way to configure the required
attributes so they only show up if you want them?

theninj4:
Thanks :)

[...]

PR #130, opened 2016-05-21
Swagger Improvements

theninj4:
Addresses #129 - We had a few errors in our dynamically generated `swagger.json` which have now been fixed.

I've added an extra CI build step to use the [swagger-tools](https://www.npmjs.com/package/swagger-tools) module to validate the generated swagger doc at build time.

I'll address the issues inline with the diff.


lib/swagger/paths.js
```diff
@@ -116,7 +129,7 @@ swaggerPaths._getPathOperationObject = function(options) {
         description: options.resourceName + " " + options.handler + " response",
         schema: {
           type: "object",
-          required: [ "jsonapi", "meta, links" ],
+          required: [ "jsonapi", "meta", "links" ],
```
theninj4: This was originally a typo... Oops?

lib/swagger/paths.js
```diff
@@ -164,6 +177,7 @@ swaggerPaths._getPathOperationObject = function(options) {
   };
   if (options.extraTags) {
     pathDefinition.tags = pathDefinition.tags.concat(options.extraTags);
+    pathDefinition.tags = _.uniq(pathDefinition.tags);
```
theninj4: This ensures all tags are unique and prevents us from having `[ "a", "a" ]`.

lib/swagger/paths.js
```diff
@@ -180,7 +194,10 @@ swaggerPaths._getPathOperationObject = function(options) {
   if (((options.handler === "search") || (options.handler === "find")) && !options.relation) {
     pathDefinition.parameters = pathDefinition.parameters.concat(swaggerPaths._optionalJsonApiParameters());
     responseShortcut.included = {
-      type: "array"
+      type: "array",
+      items: {
+        type: "object"
+      }
```
theninj4: This is the definition for what should appear in the `included` block on any request supporting inclusions. I've set it to simply be `an object` for now - it seems like overkill to say `this could be any resource in the swagger doc`.

lib/swagger/paths.js
```diff
@@ -214,7 +231,7 @@ swaggerPaths._getPathOperationObject = function(options) {
     pathDefinition.responses["200"] = undefined;
   }
 
-  if ((options.handler !== "search") && (options.handler !== "create")) {
+  if (options.hasPathId) {
```
theninj4: When we create the path operation objects, we now pass in `hasPathId` which reflects if the URL contains a `/{id}/` parameter. We used to guess based on the operation, but it wasn't specific enough for all edge cases.

lib/swagger/paths.js
```diff
@@ -258,7 +275,8 @@ swaggerPaths._optionalJsonApiParameters = function() {
     { "$ref": "#/parameters/sort" },
     { "$ref": "#/parameters/include" },
     { "$ref": "#/parameters/filter" },
-    { "$ref": "#/parameters/fields" }
+    { "$ref": "#/parameters/fields" },
+    { "$ref": "#/parameters/page" }
```
theninj4: This adds the `page=` url param to optionally appear pretty much anywhere.

lib/swagger/resources.js
```diff
@@ -30,12 +30,10 @@ swaggerPaths._getResourceDefinition = function(resourceConfig) {
       },
       "attributes": {
         type: "object",
-        required: [ ],
         properties: { }
       },
       "relationships": {
         type: "object",
-        required: [ ],
```
theninj4: We're not allowed to have empty `required` arrays.

test/get-resource-id.js
```diff
@@ -38,7 +39,7 @@ describe("Testing jsonapi-server", function() {
 
     it("with fields", function(done) {
       var url = "http://localhost:16006/rest/articles/de305d54-75b4-431b-adb2-eb6b9e546014?fields[articles]=title";
-      helpers.request({
+      request({
```
theninj4: `helpers.request` is a wrapper around `request` that also validates the request + response against the swagger definition. The `fields` JSON:API attribute allows us to bypass the required attributes on the models, so I've had to opt a bunch of these tests out of those assertions.

lib/swagger/resources.js
```diff
@@ -74,9 +72,10 @@ swaggerPaths._getResourceDefinition = function(resourceConfig) {
       }
       attributeShortcut[attribute] = swaggerScheme;
 
-      // if ((joiScheme._flags || { }).presence === "required") {
-      //   resourceDefinition.properties.attributes.required.push(attribute);
-      // }
+      if ((joiScheme._flags || { }).presence === "required") {
+        resourceDefinition.properties.attributes.required = resourceDefinition.properties.attributes.required || [ ];
+        resourceDefinition.properties.attributes.required.push(attribute);
+      }
```
theninj4: This block used to give us troubles with our test suite, especially around queries that use the `fields=` parameter to opt-out of seeing required attributes. I've fixed the tests and this can go back in. Required fields now appear correctly in the swagger doc.

lib/swagger/paths.js
```diff
@@ -40,31 +43,36 @@ swaggerPaths._addBasicPaths = function(paths, resourceName, resourceConfig) {
       handler: "search",
       resourceName: resourceName,
       description: "Search for " + resourceName,
-      parameters: resourceConfig.searchParams
+      parameters: resourceConfig.searchParams,
+      hasPathId: false
```
theninj4: All of these `hasPathId` params correspond to if the URL contains `/{id}/`.

theninj4:
If you click through to the CI build you'll see all our tasks, including the new one for swagger:

https://travis-ci.org/holidayextras/jsonapi-server/builds/131891216

pmcnr-hx:
LGTM: :+1:

connormeredith:
👍

PR #131, opened 2016-05-22
[RFC] Parse and Joi validate filter

pmcnr-hx:
Parses and Joi validate, setting values to correct types, during Joi validates.

This was a bit more invasive than initially anticipated, but it should improve the validation of filters while providing a fully parsed filter to both post-processing and handlers, while avoiding duplicated functionality for this in each handler.

- [x] :+1: :one: 
- [x] :+1: :two:  
- [x] :+1: :three:

### DIFF
--- a/lib/filter.js
+++ b/lib/filter.js
@@ -1,0 +1,31 @@
+/* @flow weak */
+"use strict";
+var filter = module.exports = { };
+
+
+filter.validateParams = function(request, callback) {
+  if (!request.params.filter) return callback();
+
+  for (var i in request.params.filter) {
+    if (request.params.filter[i] instanceof Object) continue;
+    if (!request.resourceConfig.attributes[i]) {
+      return callback({
+        status: "403",
+        code: "EFORBIDDEN",
+        title: "Invalid filter",
+        detail: request.resourceConfig.resource + " do not have property " + i
+      });
+    }
+    var relationSettings = request.resourceConfig.attributes[i]._settings;
+    if (relationSettings && relationSettings.__as) {
+      return callback({
+        status: "403",
+        code: "EFORBIDDEN",
+        title: "Request validation failed",
+        detail: "Requested relation \"" + i + "\" is a foreign reference and does not exist on " + request.params.type
+      });
+    }
+  }
+
+  return callback();
+};
--- a/lib/filter.js
+++ b/lib/filter.js
@@ -108,7 +108,7 @@ filter._parseFilterElement = function(attributeName, attributeConfig, filterElem
   return { result: helperResult.result };
 };
 
-filter.validateParams = function(request, callback) {
+filter.parseAndValidate = function(request, callback) {
   if (!request.params.filter) return callback();
 
   var resourceConfig = request.resourceConfig;
--- a/lib/filter.js
+++ b/lib/filter.js
@@ -29,7 +29,9 @@ filter.validateParams = function(request, callback) {
 
   var error;
   for (var key in request.params.filter) {
-    if (request.params.filter[key] instanceof Object) continue;
+    var filterElement = request.params.filter[key];
+
+    if (!Array.isArray(filterElement) && filterElement instanceof Object) continue;
 
     error = filter._resourceDoesNotHaveProperty(request.resourceConfig, key);
     if (error) return callback(error);
--- a/lib/filter.js
+++ b/lib/filter.js
@@ -3,28 +3,39 @@
 var filter = module.exports = { };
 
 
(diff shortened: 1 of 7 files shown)


theninj4:
Thanks for the contribution :+1: I'm sat here thinking if mutating the original `filter` object is the right thing to do. I think it'd be better for everyone else if we simply build a new object called something like `processedFilter`, `validatedFilter`, `safeFilter` or somerthing, which exists alongside the original set of filters? That way we won't be breaking any existing integrations and we still provide the nicer interface moving forwards.

pmcnr-hx:
You're sat there thinking very wise thoughts. I concur with you 100% and I will adjust accordingly. That will actually the changes less intrusive and I will be able to simplify the patch. Thanks!

lib/postProcess.js
```diff
@@ -47,9 +47,14 @@ postProcess._fetchRelatedResources = function(request, mainResource, callback) {
...
-    ids = urlJoiner + ids.join(urlJoiner);
-    return jsonApi._apiConfig.pathPrefix + type + "/?" + ids;
+    var qsField = "filter[id]=";
+    var urlJoiner = "&" + qsField;
+    ids = qsField + ids.join(urlJoiner);
+    var uri = jsonApi._apiConfig.pathPrefix + type + "/?" + ids;
+    if (request.route.query) {
+      uri += "&" + request.route.query;
+    }
+    return uri;
```
theninj4: Is this change designed to try and remove the leading `&` from the query string? It's quite hard to read by comparison, I'd rather we either live with the leading ampersand or we simply substring it off with a comment saying `// trim the leading ampersand`.

pmcnr-hx: It wasn't just that, it was because we weren't passing the filter to the included resources rerouted request (see `54-56`). With the new strategy of not altering the original filter and adding the parsed filter as a new property we won't need to touch `lib/postProcess.js` at all so I am going to remove these changes.

theninj4: Ahhh right, cool :+1:

pmcnr-hx: I've changed it back to how it was before but we still need to pass the query string to the related resources. I am not sure how the test was working before... :confounded:

pmcnr-hx:
@theninj4: implemented the proposed changes. Adding the processed filter as a new property of the request, instead if mutating the existing filter is indeed a cleaner, less invasive and backward compatible approach. :heart_eyes: Post-processing filtering is now nicely simplified, and the parsed filter is made available to the handlers simplifying their job and avoiding code duplication there.

lib/filter.js
```diff
@@ -0,0 +1,141 @@
...
+        code: "EFORBIDDEN",
+        title: "Invalid filter",
+        detail: "Filter value for key '" + attributeName + "' is invalid: " + helperResult.error
+      }
+    };
+  }
+  return { result: helperResult.result };
+};
+
+filter.process = function(request, callback) {
```
theninj4: Is `filter.process` the right name for this? Reading further down and seeing it alongside other high level features left me a bit confused as to what processing it was doing - maybe something like `filter.validateAndSanitise` could be more appropriate?

lib/filter.js
```diff
@@ -0,0 +1,141 @@
...
+
+    parsedFilterElement = filter._parseFilterElement(key, resourceConfig.attributes[key], filterElement);
+    if (parsedFilterElement.error) return callback(parsedFilterElement.error);
+
+    processedFilter[key] = [].concat(parsedFilterElement.result);
+  }
+
+  request.processedFilter = processedFilter;
+
+  return callback();
```
theninj4: Does this need to be async? We could have it return an error (if any) and default to returning null?

theninj4:
This is looking good :+1:

lib/filter.js
pmcnr-hx: Agreed. I guess the refactoring from its origins as a waterfall step led in this direction. I'll make the changes.

lib/filter.js
pmcnr-hx: Weird, I though I had renamed it to `validateAndParse`...

pmcnr-hx:
Feedback actioned @theninj4. Thanks for the +1. :grinning:

reecemillsom:
Taking 1️⃣

reecemillsom:
Looking epic 👍

elliottcrush:
🔎

elliottcrush:
Happy with the changes and actioned updates. Looks like some nice work here 👍

theninj4:
Hmm... this merged into the wrong branch :'(

pmcnr-hx:
It was! Just fixing it now.

Issue #141, opened 2016-05-31
Unable to patch relationships on resources with required attributes


BryceCicada:
JSON:API spec for patching relationships is described at http://jsonapi.org/format/#crud-updating-relationships.

> **Updating To-Many Relationships**
> And the following request clears every tag for an article:
>
    PATCH /articles/1/relationships/tags HTTP/1.1
    Content-Type: application/vnd.api+json
    Accept: application/vnd.api+json           
    {
      "data": []
    }

However, attempting this on resources with required attributes results in HTTP 403 Forbidden with detail "Param validation failed".  

This can be reproduced on the example project with:

    PATCH http://localhost:16006/rest/articles/de305d54-75b4-431b-adb2-eb6b9e546014/relationships/tags
    Content-Type: application/vnd.api+json
    Accept: application/vnd.api+json           
    {
      "data": []
    }

which results in:

    Status Code: 403 Forbidden
    {
      "jsonapi": {
        "version": "1.0"
      },
      "meta": {
        "description": "This block shows up in the root node of every payload"
      },
      "links": {
        "self": "http://localhost:16006/rest/articles/de305d54-75b4-431b-adb2-eb6b9e546014/relationships/tags"
      },
      "errors": [
        {
          "status": "403",
          "code": "EFORBIDDEN",
          "title": "Param validation failed",
          "detail": [
            {
              "message": "\"title\" is required",
              "path": "title",
              "type": "any.required",
              "context": {
                "key": "title"
              }
            },
            {
              "message": "\"content\" is required",
              "path": "content",
              "type": "any.required",
              "context": {
                "key": "content"
              }
            }
          ]
        }
      ]

theninj4:
Fixed in v1.11.0

PR #145, opened 2016-06-02
Unable to patch relationships on resources with required attributes

theninj4:
Addresses #141 - When patching a resources relationship, we should only be validating the named relationship that is being mutated. I've adjusted the example server to reflect the bug, fixed the issue, and the test suite is passing again.


pmcnr-hx:
LGTM: :+1:

Issue #142, opened 2016-05-31
Null pointer


jbockerstette:
/root/jsonapi-server/lib/routes/relationships.js:47
        sanitisedData = sanitisedData.relationships[request.params.relation].data;
                                     ^

TypeError: Cannot read property 'relationships' of null
    at /root/jsonapi-server/lib/routes/relationships.js:47:38
    at fn (/root/jsonapi-server/node_modules/async/lib/async.js:746:34)
    at /root/jsonapi-server/node_modules/async/lib/async.js:1213:16
    at /root/jsonapi-server/node_modules/async/lib/async.js:166:37
    at /root/jsonapi-server/node_modules/async/lib/async.js:706:43
    at /root/jsonapi-server/node_modules/async/lib/async.js:167:37
    at Immediate._onImmediate (/root/jsonapi-server/node_modules/async/lib/async.js:                                              1206:34)
    at tryOnImmediate (timers.js:534:15)
    at processImmediate [as _immediateCallback] (timers.js:514:5)

theninj4:
This occurs when the data coming out of your handler doesn't validate against your schema.

For example, if your Joi schema for a resource says `There will be a "body" attribute`, but the data in a database somewhere doesn't enforce the same requirement and someone inserts a row without a `body`, then you try and fetch that resource via the API.

I'm opening a PR in to provide an error saying "Sorry, this resource exists, but its fundamentally broken. We can't provide you with the resource because it won't conform to your expectations."

theninj4:
Fixed in v1.11.0

PR #146, opened 2016-06-02
Error when interacting with broken resources

theninj4:
This addresses #142 - If we pull a resource out of a handler and it doesn't validate, we won't respond with it. In that case, we should respond with a sensible error message saying "Sorry, this resource exists, but its fundamentally broken. We can't provide you with the resource because it won't conform to your expectations."


pmcnr-hx:
LGTM: :+1:

PR #148, opened 2016-06-02
Allow some methods to be omitted from generated swagger docs

BryceCicada:
I have custom handlers that do not allow some operations on resources and/or relations.  I'd like the generated swagger docs to reflect this.

Here's my proposal.


adamlc:
Interestingly I have certain methods disabled in my adapter disabled, i.e.

```nodejs

module.exports = new jsonApi.MemoryHandler();
module.exports.create = null;
module.exports.delete = null;
```

Shouldn't the swagger generator also pick those up? As it seems to put them in anyway :(

pmcnr-hx:
@adamlc suggestion does appeal and it seems it would be simpler to implement. What do you think @BryceCicada: only add method to the swagger docs if it is defined in the handler?

hpoom:
👍 to @adamlc implementation. I like that it is very clear in the code that those methods are not supported, instead of just forgotten about.

theninj4:
The issue @BryceCicada was trying to solve here is the Swagger documentation of the individual JSON:API "routes" for mutating specific relationships. There are situations whereby we have a resource that can be updated, however it has one or more relationships that aren't modifiable by the API consumer - read only relationships. Right now, if a handler exposes update functionality, the Swagger doc generator will assume all relationships are open to modifications and will generate all the possible documentation. 

I've been sat on the fence with this change - I'm not convinced Swagger is the right tool for the job for documenting a JSON:API service, mainly because of issues like this one. I'm also undecided if jsonapi-server should validate attempts to update relationships via the same config as the docs. Basically I'm waiting until this becomes a problem for us in production, hoping that the experiences we're gaining on the journey lead us to an elegant solution.

adamlc:
Makes sense I guess, as regards my comment shall I open another issue? I guess thats a little different to what @BryceCicada wants to achieve.

pmcnr-hx:
@adamlc since the scope of your issue seems to be different, it's probably best to open a new issue.

adamlc:
No problem, opened an issue in #190

PR #150, opened 2016-06-07
Clone all objects passing in-to and out-of the in memory store

theninj4:
We've had some issues with stubbing journey tests via the memory handler, and they boil down to how the memory handler doesn't clone objects passing through it. This PR ensures all objects get cloned to allow us to mutate resources without effecting the data in the store.

### DIFF
--- a/lib/MemoryHandler.js
+++ b/lib/MemoryHandler.js
@@ -36,7 +36,7 @@ MemoryStore.prototype.search = function(request, callback) {
   if (request.params.page) {
     results = results.slice(request.params.page.offset, request.params.page.offset + request.params.page.limit);
   }
-  return callback(null, this._clone(results), resultCount);
+  return callback(null, MemoryStore._clone(results), resultCount);
 };
 
 /**
@@ -59,7 +59,7 @@ MemoryStore.prototype.find = function(request, callback) {
   }
 
   // Return the requested resource
-  return callback(null, this._clone(theResource));
+  return callback(null, MemoryStore._clone(theResource));
 };
 
 /**
@@ -67,10 +67,8 @@ MemoryStore.prototype.find = function(request, callback) {
  */
 MemoryStore.prototype.create = function(request, newResource, callback) {
   // Check to see if the ID already exists
-  var matches = resources[request.params.type].filter(function(resource) {
-    return resource.id === newResource.id;
-  }).length;
-  if (matches > 0) {
+  var index = MemoryStore._indexOf(resources[request.params.type], newResource);
+  if (index !== -1) {
     return callback({
       status: "403",
       code: "EFORBIDDEN",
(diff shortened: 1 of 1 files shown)


connormeredith:
Taking review ☝️

lib/MemoryHandler.js
```diff
@@ -133,3 +157,7 @@ MemoryStore.prototype._sortList = function(request, list) {
     }
   });
 };
+
+MemoryStore.prototype._clone = function(obj) {
+  return JSON.parse(JSON.stringify(obj));
```
pmcnr-hx: Are we sure we never have circular references?

theninj4: This data is intended to be serialised at this point for storage. If anything does have a circular reference at this point, it's going to break for real when the memory handler gets replaced for a real storage mechanism?

pmcnr-hx: Since we don't really need the `MemoryStore` object instance here, I am wondering if this shouldn't simply me a module level helper function, i.e., `MemoryStore._clone`.

lib/MemoryHandler.js
```diff
@@ -80,9 +92,14 @@ MemoryStore.prototype.delete = function(request, callback) {
...
-    var resourceIndex = resources[request.params.type].indexOf(theResource);
-    resources[request.params.type].splice(resourceIndex, 1);
+    // Remove the resource from the in-memory store.
+    resources[request.params.type].some(function(resource, index) {
+      var match = resource.id === theResource.id;
+      if (match) {
+        resources[request.params.type].splice(index, 1);
+      }
+      return match;
+    });
```
pmcnr-hx: I shall note the clever use of `Array.some` while at the same time lamenting the regrettable side effect in its callback. Since [`Array.prototype.findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) is ES6 only I guess this is as good as the alternative approaches.

lib/MemoryHandler.js
```diff
@@ -59,17 +59,29 @@ MemoryStore.prototype.find = function(request, callback) {
...
 
 /**
-  Create (store) a new resource give a resource type and an object.
+  Create (store) a new resource given a resource type and an object.
  */
 MemoryStore.prototype.create = function(request, newResource, callback) {
+  // Check to see if the ID already exists
+  var matches = resources[request.params.type].filter(function(resource) {
+    return resource.id === newResource.id;
+  }).length;
```
pmcnr-hx: Now **here** would be a totally legitimate, functional and more efficient (than `.filter`) use of `Array.prototype.some`.

lib/MemoryHandler.js
```diff
@@ -133,3 +144,14 @@ MemoryStore.prototype._sortList = function(request, list) {
     }
   });
 };
+
+MemoryStore._clone = function(obj) {
```
connormeredith: Any reason for not using lodash's `cloneDeep`?

connormeredith:
Minor comment, otherwise it's looking good 👍

lib/MemoryHandler.js
theninj4: It'd mean adding another dependency for something we really don't need.

pmcnr-hx:
:+1: :shipit: I'm so happy you've implemented that `MemoryStore._indexOf` helper function. 🦄

Issue #143, opened 2016-06-01
Access to inner server


vnen:
There seems to be no way of accessing the inner server of this module. Even to use an existing express instance needs to mess with the `require` cache (as per the example in the docs).

I particularly want to run a Websockets server in the same host and port, which would be possible if either I had access to the inner server or better yet: if I could specify a server in `setConfig()`.

BryceCicada:
+1

PR #153, opened 2016-06-16
Expose ExpressJS server

theninj4:
Addresses #143. This PR enables us to interact with the underlying Express application before jsonapi-server touches it:

```javascript
  var app = jsonApi.getExpressServer();
  app.use(someMiddleware);
  jsonApi.start() // this line applies the json:api routing and starts the service
```

You can see this in action by changing the `before` block in `test/404.js` to something like this:
```
before(function() {
  jsonApiTestServer.getExpressServer().use(function(req, res, next) {
    console.log("Middleware just ran");
    return next();
  });
  jsonApiTestServer.start();
});
```
...and then run the test suite.

This PR also opens up the possibility of passing an existing ExpressJS app into jsonapi-server.

### DIFF
--- a/documentation/configuring.md
+++ b/documentation/configuring.md
@@ -23,7 +23,7 @@ jsonApi.setConfig({
 
 To run over HTTPS, set the protocol to _https_ and configure the appropriate TLS settings
 
-For example: 
+For example:
 
 ```javascript
 var fs = require("fs");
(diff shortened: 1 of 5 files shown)


theninj4:
The whitespace diff makes this way easier to review: https://github.com/holidayextras/jsonapi-server/pull/153/files?w=1

pmcnr-hx:
Useful! 👍

Issue #144, opened 2016-06-01
POST a new relationship for a Joi.one("people") in articles fails with 403


jbockerstette:
If you use your example project, do the following:
1. POST a new article
2. POST a new relationship to author of type people to the new article. Here is the curl.
`curl -X POST --header 'Content-Type: application/vnd.api+json' --header 'Accept: application/vnd.api+json' -d '{
  "data": {
    "type": "people",
    "id": "cc5cca2e-0dd8-4b95-8cfc-a11230e73116",
    "meta": {}
  }
}' 'http://localhost:16006/rest/articles/41748e1e-ecfe-47a3-b7d8-11a089420cc2/relationships/author'`

3. It will fail with 403. It appears that it does not know the type based on the null in the details messsage.

{
  "jsonapi": {
    "version": "1.0"
  },
  "meta": {
    "description": "This block shows up in the root node of every payload"
  },
  "links": {
    "self": "http://localhost:16006/rest/articles/41748e1e-ecfe-47a3-b7d8-11a089420cc2/relationships/author"
  },
  "errors": [
    {
      "status": "403",
      "code": "EFORBIDDEN",
      "title": "Param validation failed",
      "detail": [
        {
          **"message": "\"author\" must be one of [null]",**
          "path": "author",
          "type": "any.allowOnly",
          "context": {
            "valids": [
              null
            ],
            "key": "author"
          }
        },
        {
          "message": "\"author\" must be an object",
          "path": "author",
          "type": "object.base",
          "context": {
            "key": "author"
          }
        }
      ]
    }
  ]
}

jbockerstette:
The problem is in lib/routes/addRelation.js
Here is the line. This works as long as the relation is an array relationship but if it is a Joi.one relationship, then you cannot use this line.
`theirResource[request.params.relation] = theirResource[request.params.relation] || [ ];`

Here is how the logic should work:

```
if (relation is type array == true) {
    //It is an array
    theirResource[request.params.relation] = theirResource[request.params.relation] || [ ];
    theirResource[request.params.relation].push(theirs);
} else {
    //It is a Joi.one
    theirResource[request.params.relation] = theirs;
}
```

theninj4:
:+1: Thanks for reporting this. I'm working on reproducing this in a test, a fix will be coming shortly after.

jbockerstette:
I already fixed it. I will submit a pull request.

jbockerstette:
[...]

PR #154, opened 2016-06-21
Mutating singular relationships via direct http route

theninj4:
Addresses #144. When mutating a resources relationships on a `hasOne` relationship currently breaks in master. I've implemented the fix suggested by @jbockerstette and I've added tests to prove it now works for both POST'ing and DELETE'ing to both `hasOne` and `hasMany` relationships.


pmcnr-hx:
😄 👍

PR #158, opened 2016-06-28
Scrap sub-milliscond accuracy from metrics

theninj4:
Sub-millisecond accuracy seems to lead to inconsistent readings. Lets go back to using regular `Date()`s.


pmcnr-hx:
Let's do it 👍 :shipit:.

I'm still curious to look into the reason for this, but Bunyan for example uses `Date` so there's probably no compelling reason for high resolution timers when millisecond accuracy is sufficient.

Issue #161, opened 2016-06-29
Optional bypassing of post-processing filtering


pmcnr-hx:
Look into allowing post-processing filtering to be optionally bypassed so data store side filtering is properly exercised by the relevant tests.

theninj4:
We should be able to validate this by comparing the pagination block with the number of resulting resources. The post processing filter will remove resources but it will NOT twiddle the pagination counters (one of them is the claimed number of returned resources). If the two sources don't line up, the data store isn't filtering correctly.

pmcnr-hx:
Sounds like a plan. I'll see about adding a test for this to avoid future regressions.

PR #163, opened 2016-06-30
Test warnings when handler doesnt filter resources

theninj4:
I've added to the test helper so every request goes through some basic pagination validation.

The goal is to use the pagination block to work out how many resources the data handler *thinks* should be present in the response, then compare that with however many resources appear in the resulting payload. If the two differ, the handler isn't properly implementing filtering.

Right now the in-memory-handler used by the test suite for this project doesn't implement filtering (we rely on the post processing to clean it up). As such, running the test suit should show a whole bunch of warnings:
```
  Testing jsonapi-server
    Searching for resources
      applying filter
        ✓ value of wrong type should error
!!!!!!!!!!!!
http://localhost:16006/rest/articles?filter[title]=How%20to%20AWS
WARNING: Pagination count doesn't match resource count.
This usually indicates the resource hanlder is not filtering correctly!
!!!!!!!!!!!!
        ✓ equality for strings
!!!!!!!!!!!!
http://localhost:16006/rest/articles?filter[views]=10
WARNING: Pagination count doesn't match resource count.
This usually indicates the resource hanlder is not filtering correctly!
!!!!!!!!!!!!
```

Within this project, those warnings are fine and expected. Nobody should ever be using the in-memory-handler, it simply needs to validate the rest of the project functions correctly. If any handlers use this test suite to validate their functionality (which is surprisingly effective at proving all the functionality is covered) then seeing these warnings is bad and it means something's broken.

### DIFF
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -56,7 +56,31 @@ testHelpers.validateJson = function(json) {
   assert.ok(json.links instanceof Object, "Response should have a links block");
   assert.ok(!(json.errors instanceof Object), "Response should not have any errors");
   assert.equal(typeof json.links.self, "string", "Response should have a \"self\" link");
+  testHelpers.validatePagination(json);
   return json;
+};
+
+testHelpers.validatePagination = function(json) {
+  if (!json.meta.page) return;
+  if (!(json.data instanceof Array)) return;
+
+  var page = json.meta.page;
+  var expectedCount = null;
+  if ((page.offset + page.limit) > page.total) {
+    expectedCount = page.total - page.offset;
+  } else if (page.limit < page.total) {
+    expectedCount = page.limit;
+  } else {
+    expectedCount = page.total;
+  }
+
+  if (expectedCount !== json.data.length) {
+    console.warn("!!!!!!!!!!!!");
+    console.warn(json.links.self);
+    console.warn("WARNING: Pagination count doesn't match resource count.");
+    console.warn("This usually indicates the resource hanlder is not filtering correctly!");
+    console.warn("!!!!!!!!!!!!");
+  }
 };
 
 testHelpers.validateRelationship = function(relationship) {


theninj4:
Helps #161

test/helpers.js
```diff
@@ -56,9 +56,33 @@ testHelpers.validateJson = function(json) {
   assert.ok(json.links instanceof Object, "Response should have a links block");
   assert.ok(!(json.errors instanceof Object), "Response should not have any errors");
   assert.equal(typeof json.links.self, "string", "Response should have a \"self\" link");
+  testHelpers.validatePagination(json);
   return json;
 };
 
+testHelpers.validatePagination = function(json) {
+  if (!json.meta.page) return;
+  if (!(json.data instanceof Array)) return;
```
pmcnr-hx: `Array.isArray` seems to be recommend when supported (as in Node's case) since it covers some corner cases that `instanceof Array` doesn't.

test/helpers.js
```diff
@@ -56,9 +56,33 @@ testHelpers.validateJson = function(json) {
...
+
+  var page = json.meta.page;
+  var expectedCount = null;
+  if ((page.offset + page.limit) > page.total) {
```
pmcnr-hx: I personally wouldn't bother with the redundant `()`. Do you reckon it deters from readability?

pmcnr-hx:
Just minor non-blocker comments which I will leave at your discretion: 👍 :shipit:

test/helpers.js
theninj4: The rest of this project uses `instanceof Array`, so we should stick with it.

test/helpers.js
theninj4: I see the brackets as an important part of identifying the left hand side from the right hand side.

PR #165, opened 2016-07-05
Delete one relationship

### LINKED ISSUE #164: Deleting one-to-one relationship fails with error
Hi!

```
TypeError: theirResource[request.params.relation].map is not a function
    at node_modules/jsonapi-server/lib/routes/removeRelation.js:43:59
```
The problem is here:
```
var keys = theirResource[request.params.relation].map(function(j) {
          return j.id;
        });
```
For one-to-one relation  ` theirResource[request.params.relation]` is not an array.
And I could not find any tests covering this case.

theninj4:
Fixes #164. I thought we'd squashed this one in #154 but it looks like one of the tests wasn't quite right and thus wasn't exercising the code properly.

This PR also includes some additional test suite warnings for use with other data handlers to warn if the data layer filtering isn't functioning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- 2016-07-05 - v1.15.1
+- 2016-07-05 - Fixed bug when deleting singular relationships via deep urls
 - 2016-06-28 - v1.15.0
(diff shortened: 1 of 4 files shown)


pmcnr-hx:
Assuming it will be rebased on master: 👍 :shipit:

PR #166, opened 2016-07-05
Updating eslintrc following major version bump

theninj4:
When we last bumped the version of eslint and jumped major versions we forgot to opt-in to the recommended ruleset. This PR fixes up our config file to ensure we get the most value out of eslint.


pmcnr-hx:
👍 :shipit:

PR #169, opened 2016-07-06
Prevent express from appending the charset to the content-type header

theninj4:
Express has some interesting behaviour whereby when calling `.send` (or equivalent) with a string, it will override the Content-Type header to append a `charset`:
https://github.com/expressjs/express/issues/2238

This is giving us trouble in our client module whereby `superagent` copies the content-type of the server and replays it back to us. The content type we're currently outputting (`application/vnd.api+json; charset=UTF-8`) isn't one that we accept back - the JSON:API spec specifically states `Servers MUST respond with a 415 Unsupported Media Type status code if a request specifies the header Content-Type: application/vnd.api+json with any media type parameters.`.

This PR changes things so we give express a `Buffer` which bypasses their twiddling of our content-type header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- 2016-07-06 - v1.15.4
+- 2016-07-06 - Prevent express from appending the charset to the content-type header
 - 2016-07-06 - v1.15.3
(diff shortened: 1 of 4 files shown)


pmcnr-hx:
Fascinating! 👍 🚢 it.

PR #173, opened 2016-08-04
Fix memory handler pagination

pmcnr-hx:
Fix memory handler pagination.

The memory handler wasn't calculating the slice limits correctly, and our test actually tested for the erroneous behaviour: fix the test and correct the calculation of the slice limits.

- [ ]  Me gusta! 👍


pmcnr-hx:
Misunderstood the behaviour. Previous behaviour is the correct one. It's resource offset, **not** page offset.

PR #175, opened 2016-08-30
allow resourceHandler.initiali'z'e

rynomad:
just a quick assignment in case someone uses the 'z' spelling... would have saved me a few minutes today.


pmcnr-hx:
Thank you for your contribution @rynomad. I guess the American English spelling will be the most familiar for most users. I believe there's some documentation that will also need to be updated to mention this alias, so I'll just check that before merging.

Issue #176, opened 2016-08-31
Event listeners / pre or post processing for resources.


adamlc:
We're just starting to build a new API around jsonapi-server and loving it so far! 

I just had a query / suggestion. We have a requirement where we need to do various things either before or after a resource is created. The main requirement is we need to quota certain resources so we'd need to hook in before records are created and after they're created. Returning an error if their quota is full or adding to their quota once a record is inserted.

At the moment the only way of doing this from what I can see if using a Custom Handler. Whilst this would work it would mean creating a Custom Handler for each resource that needs a hook with only subtle differences.

What are people's thoughts on having a common way to hook into these events? It would be great if this was built into the core so each handler didn't need to be rebuilt. Sorry for the waffle, I hope it makes sense!

theninj4:
This has come up before in #162, and we ran in to a challenge in production a few months back whereby we wanted to add some custom authentication to a bunch of resources backed by the relational data store, we solved it by creating a custom handler that accepts another handler and allows us to inject custom functionality around each handler action.

I'll look in to this issue today and see if I can find a syntactically pleasant solution that can be open sourced.

theninj4:
@adamlc does the above PR help overcome some of your challenges?

adamlc:
@theninj4 that looks like it should do the trick! I'll have a play later and let you know how it goes :)

adamlc:
@theninj4 Just to let you know that worked perfectly for our use case, can't thank you enough!

PR #178, opened 2016-09-06
Exploring the possibility of a ChainHandler

theninj4:
This is a work in progress and a proof of concept for a possible solution to #176 the idea being we can chain together handlers:

### Chaining handlers together with the ChainHandler

This project ships with a handler called the `ChainHandler`. It's purpose is to make it easier to wrap the functionality of one handler with additional functionality. For example using the relational data store may be ideal for storing resources, but what if we want to add an authentication layer on top of those resources? An authentication handler shouldn't know or care about the storage of the underlying resource, it should only deal with authentication - we need an abstraction layer to help us glue all of this together. 

To get started with the ChainHandler, go ahead and instantiate one:
```javascript
var testChainHandler = new jsonApi.ChainHandler();
```
From there it's a simple case of attaching new function handlers to `testChainHandler`, following the naming convention of `("before"|"after")+function` - eg `beforeSearch`, `beforeCreate`, `afterDelete`.

These new functions need to match the signatures of the functions they're overriding and need to pass the data onwards via a callback. For example, here are two valid functions that don't add any functionality. Notice how the callback values match the function signatures:
```javascript
testChainHandler.beforeSearch = function(request, callback) {
  return callback(null, request);
};

testChainHandler.afterSearch = function(results, pagination, callback) {
  return callback(null, results, pagination);
};
```

With a new ChainHandler ready to go, this is how we can stack them up:
```javascript
jsonApi.define({
  resource: "articles",
  handlers: chainHandler1.chain(chainHandler2).chain(memoryHandler),
});
```

The call stack when chaining multiple `ChainHandler`s together looks like this:
```
chainHandler1.beforeSearch
  chainHandler2.beforeSearch
    MemoryHandler.search
  chainHandler2.afterSearch
chainHandler1.afterSearch
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Ultimately, the only things you as a user of this framework need to care about a
 
 We've created `handler`s to automatically map our config over to database solutions help people get off the ground:
  * [jsonapi-store-memoryhandler](https://github.com/holidayextras/jsonapi-server/blob/master/lib/MemoryHandler.js) - an in-memory data store to enable rapid prototyping. This ships as a part of `jsonapi-server` and powers the core test suite.
- * [jsonapi-store-chainhandler](https://github.com/holidayextras/jsonapi-server/blob/master/lib/ChainHandler.js) - a handler to chain custom behaviour around an existing handler. This ships as a part of `jsonapi-server`.
+ * [jsonapi-handler-chain](https://github.com/holidayextras/jsonapi-server/blob/master/lib/ChainHandler.js) - a handler to chain custom behaviour around an existing handler. This ships as a part of `jsonapi-server`. More info can be found [here](https://github.com/holidayextras/jsonapi-server/blob/master/documentation/chain-handler.js)
  * [jsonapi-store-relationaldb](https://github.com/holidayextras/jsonapi-store-relationaldb) - using `sequelize` to support PostgreSQL, MySQL, MSSQL, MariaDB and SQLite.
  * [jsonapi-store-mongodb](https://github.com/holidayextras/jsonapi-store-mongodb) - for MongoDB.
  * [jsonapi-store-elasticsearch](https://github.com/holidayextras/jsonapi-store-elasticsearch) - for Elasticsearch.
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Ultimately, the only things you as a user of this framework need to care about a
 
 We've created `handler`s to automatically map our config over to database solutions help people get off the ground:
  * [jsonapi-store-memoryhandler](https://github.com/holidayextras/jsonapi-server/blob/master/lib/MemoryHandler.js) - an in-memory data store to enable rapid prototyping. This ships as a part of `jsonapi-server` and powers the core test suite.
+ * [jsonapi-store-chainhandler](https://github.com/holidayextras/jsonapi-server/blob/master/lib/ChainHandler.js) - a handler to chain custom behaviour around an existing handler. This ships as a part of `jsonapi-server`.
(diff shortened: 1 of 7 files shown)


lib/ChainHandler.js
```diff
@@ -0,0 +1,43 @@
...
+
+var ChainHandler = module.exports = function ChainHandler() {
+};
+
+ChainHandler.prototype.chain = function(otherHandler) {
+  var self = this;
+  self.otherHandler = otherHandler;
+  self.ready = true
+  return self;
+};
```
pmcnr-hx: Couldn't we just move the `chain` functionality to the constructor function?

README.md
```diff
@@ -34,6 +34,7 @@ Ultimately, the only things you as a user of this framework need to care about a
 
 We've created `handler`s to automatically map our config over to database solutions help people get off the ground:
  * [jsonapi-store-memoryhandler](https://github.com/holidayextras/jsonapi-server/blob/master/lib/MemoryHandler.js) - an in-memory data store to enable rapid prototyping. This ships as a part of `jsonapi-server` and powers the core test suite.
+ * [jsonapi-store-chainhandler](https://github.com/holidayextras/jsonapi-server/blob/master/lib/ChainHandler.js) - a handler to chain custom behaviour around an existing handler. This ships as a part of `jsonapi-server`.
```
pmcnr-hx: `jsonapi-handler-chain` instead, since it's not really a store?

pmcnr-hx:
This is marvellous! (ignore my outdated comment, having the separate `.chain` method reads much better).

README.md
```diff
@@ -34,6 +34,7 @@ Ultimately, the only things you as a user of this framework need to care about a
 
 We've created `handler`s to automatically map our config over to database solutions help people get off the ground:
  * [jsonapi-store-memoryhandler](https://github.com/holidayextras/jsonapi-server/blob/master/lib/MemoryHandler.js) - an in-memory data store to enable rapid prototyping. This ships as a part of `jsonapi-server` and powers the core test suite.
+ * [jsonapi-handler-chain](https://github.com/holidayextras/jsonapi-server/blob/master/lib/ChainHandler.js) - a handler to chain custom behaviour around an existing handler. This ships as a part of `jsonapi-server`. More info can be found [here](https://github.com/holidayextras/jsonapi-server/blob/master/documentation/chain-handler.js)
```
pmcnr-hx: `...documentation/chain-handler.js` → `...documentation/chain-handler.md`

lib/ChainHandler.js
```diff
@@ -0,0 +1,46 @@
+/* @flow weak */
+"use strict";
+
+var async = require("async");
+
+var ChainHandler = module.exports = function ChainHandler() { };
+
+ChainHandler.prototype.chain = function(otherHandler) {
```
pmcnr-hx: I'm wondering if calling the parameter `nextHandler` would make clearer the the order of invocation... Just a suggestion though, it's fine as is.

pmcnr-hx:
Excellent work, and all looks legit! Worth adding a few automated tests to cover the new functionality and avoid future regressions?

theninj4:
I've left a dual-chained handler in the example project used by the tests - if we break the ChainHandler the test suite fails catastrophically ;)

lib/ChainHandler.js
theninj4: I like the use of `chain` as when we start stacking up these handlers it won't be very obvious where the magic is coming from - by including a part of the name in the function we're helping to steer newcomers towards the right documentation.

PR #184, opened 2016-09-08
ChainHandler improvements

theninj4:
After implementing an `AuthenticationHandler` using the `ChainHandler` I ran in to two issues which I've fixed:

1. There was no way of moving data in scope of a `before` block into an `after` block. I've solved this by making the `request` object available in all `after` blocks. This enables us to implement clever features on a per-request basis by providing a mechanism to push state between `before` and `after` functions.

2. The `.initialise()` function is synchronous, and despite if any before/after blocks immediately invoke their callback's the `waterfall` module deschedules them, causing the functions to fire out of order. I've added a cleaner path through the ChainHandler for synchronous functions to ensure this doesn't happen. This change enables us to use `beforeInitialise` to add attributes to resources before they hit a storage handler, so we can magically add accessTokens to enable authentication checks.

### DIFF
--- a/documentation/chain-handler.md
+++ b/documentation/chain-handler.md
@@ -1,7 +1,7 @@
 
 ### Chaining handlers together with the ChainHandler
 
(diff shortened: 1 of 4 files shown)


connormeredith:
Looks good man, 👍  when CI passes

Issue #186, opened 2016-09-08
Mask local routing addresses for an alias


st3xupery:
When generating absolute links it would be ideal to mask the the addresses being generated to an alias that differs from the config parameters provided on server initialization. The port number I may provide to the server will not be the port number of the address someone reaches the API from. With nginx I proxy the traffic that hits the node server and I'd like the links to reflect that.

I've implemented such a change on [my fork here](https://github.com/st3xupery/jsonapi-server/tree/url-prefix-alias) with a field called urlPrefixAlias to be read at initialization. The changes to the code are minimal, doesn't interfere with existing tests. Let me know if I should write up some tests before I make a PR.

theninj4:
Your changes look good, I'm not too worried about tests (it's very hard to test with the current test suite, which is geared towards a full-stack run journey via an example server) but some additional documentation might be good? Does the swagger documentation need any minor tweaking too? In summary - looks good, would merge :+1:

st3xupery:
Great! Added some additional docs to configure.md and made sure the changes reflect in the swagger output.

theninj4:
Thanks again for the contribution! :+1:

PR #188, opened 2016-09-09
Mask local routing addresses for an alias

st3xupery:
Resolves issue #186

### DIFF
--- a/documentation/configuring.md
+++ b/documentation/configuring.md
@@ -3,6 +3,9 @@
 
 ```javascript
 jsonApi.setConfig({
+  // (optional) An alias of the absolute portion of URLs generated in a response file
+  // eg http://localhost:16006/some-resource/ -> https://www.example.com/my-api/some-resource/
+  urlPrefixAlias: "https://www.example.com/my-api/",
   // (optional) HTTP / HTTPS
   protocol: "http",
   // (optional) The hostname the API will be sat behind, from the customer's perspective
--- a/lib/jsonApi.js
+++ b/lib/jsonApi.js
@@ -26,7 +26,7 @@ jsonApi.ChainHandler = require("./ChainHandler");
 jsonApi.setConfig = function(apiConfig) {
   jsonApi._apiConfig = apiConfig;
   jsonApi._apiConfig.base = jsonApi._cleanBaseUrl(jsonApi._apiConfig.base);
-  jsonApi._apiConfig.pathPrefix = jsonApi._concatenateUrlPrefix(jsonApi._apiConfig);
+  jsonApi._apiConfig.pathPrefix = apiConfig.urlPrefixAlias || jsonApi._concatenateUrlPrefix(jsonApi._apiConfig);
   responseHelper.setBaseUrl(jsonApi._apiConfig.pathPrefix);
   responseHelper.setMetadata(jsonApi._apiConfig.meta);
 };
--- a/lib/router.js
+++ b/lib/router.js
@@ -164,9 +164,10 @@ router._getParams = function(req) {
     "host", "connection", "accept-encoding", "accept-language", "content-length"
   ];
 
-  var combined;
+  var combined, reqUrl = req.url;
   if (jsonApi._apiConfig.urlPrefixAlias) {
     combined = jsonApi._apiConfig.urlPrefixAlias.replace(/\/$/, '');
+	reqUrl = reqUrl.replace(jsonApi._apiConfig.base, '/');
   } else {
     combined = url.format({
       protocol: jsonApi._apiConfig.protocol,
@@ -175,7 +176,7 @@ router._getParams = function(req) {
     });
   }
 
-  combined += req.url;
+  combined += reqUrl;
 
   return {
     params: _.assign(req.params, req.body, req.query),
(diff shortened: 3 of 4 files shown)


theninj4:
:+1: I'm on holiday this week, first thing Monday I'll test, merge and publish :)

theninj4:
So I'm testing this by tweaking the example server (used by the test suite) to set `urlPrefixAlias` to be the same value it would otherwise default to: `"http://localhost:16006/rest/"`. This lets me test out the new code paths using the existing test suite.

It seems to be flagging up an issue when creating resources whereby we have to set a response header of `Location` to be the URL to the newly created resource, it's handled by this:
```
        res.set({
          "Location": request.route.combined + "/" + newResource.id
        });
```
With an error showing the `combined` route isn't joining correctly:
```
Uncaught AssertionError: 
  'http://localhost:16006/rest/rest/photos/150dee1d-6375-4e0b-a317-7e13ba32cabe' == 
  'http://localhost:16006/rest/photos/150dee1d-6375-4e0b-a317-7e13ba32cabe'
```

There's a similar issue with the pagination url's:
```
links: {
  self: "http://localhost:16006/rest/photos?page[limit]=1",
  last: "http://localhost:16006/rest/rest/photos?page%5Blimit%5D=1&page%5Boffset%5D=2",
  next: "http://localhost:16006/rest/rest/photos?page%5Blimit%5D=1&page%5Boffset%5D=1"
},
```

theninj4:
If I assume the `urlPrefixAlias` isn't meant to include `rest/` and omit it, I get test failures in the opposite direction:
```diff
"relationships": {
  "article": {
    "links": {
-        "related": "http://localhost:16006/articles/?filter[comments]=3f1a89c2-eb85-4799-a048-6735db24b7eb"
-        "self": "http://localhost:16006/articles/relationships/?comments=3f1a89c2-eb85-4799-a048-6735db24b7eb"
+        "related": "http://localhost:16006/rest/articles/?filter[comments]=3f1a89c2-eb85-4799-a048-6735db24b7eb"
+        "self": "http://localhost:16006/rest/articles/relationships/?comments=3f1a89c2-eb85-4799-a048-6735db24b7eb"
```

theninj4:
This changeset gets things working:
```diff
diff --git a/example/server.js b/example/server.js
index b7ddfdc..92daecb 100644
--- a/example/server.js
+++ b/example/server.js
@@ -21,6 +21,7 @@ jsonApi.setConfig({
       url: "http://opensource.org/licenses/MIT"
     }
   },
+  urlPrefixAlias: "http://localhost:16006/",
   protocol: "http",
   hostname: "localhost",
   port: 16006,
diff --git a/lib/jsonApi.js b/lib/jsonApi.js
index 1f1909b..56ee294 100644
--- a/lib/jsonApi.js
+++ b/lib/jsonApi.js
@@ -26,7 +26,7 @@ jsonApi.ChainHandler = require("./ChainHandler");
 jsonApi.setConfig = function(apiConfig) {
   jsonApi._apiConfig = apiConfig;
   jsonApi._apiConfig.base = jsonApi._cleanBaseUrl(jsonApi._apiConfig.base);
-  jsonApi._apiConfig.pathPrefix = apiConfig.urlPrefixAlias || jsonApi._concatenateUrlPrefix(jsonApi._apiConfig);
+  jsonApi._apiConfig.pathPrefix =jsonApi._concatenateUrlPrefix(jsonApi._apiConfig);
   responseHelper.setBaseUrl(jsonApi._apiConfig.pathPrefix);
   responseHelper.setMetadata(jsonApi._apiConfig.meta);
 };
@@ -34,6 +34,11 @@ jsonApi.setConfig = function(apiConfig) {
 jsonApi.authenticate = router.authenticateWith;
 
 jsonApi._concatenateUrlPrefix = function(config) {
+  if (config.urlPrefixAlias) {
+    var result = config.urlPrefixAlias + '/' + config.base;
+    result = result.replace(/([^:])\/+/g, '$1/');
+    return result;
+  }
   return url.format({
     protocol: config.protocol,
     hostname: config.hostname,
```
Do they seem reasonable @st3xupery ?

st3xupery:
> If I assume the urlPrefixAlias isn't meant to include rest/ and omit it
Correct, the urlPrefixAlias should mask the base url param as well.

I sourced the problem to the req.url value being pulled into the combined URL value as can be seen [here](https://github.com/holidayextras/jsonapi-server/commit/608e90a8cba14b3d4e02f9b91517fc3a8e2d16fe#diff-504fd89f344be1e9c4b583f3ebd7428aR167)

What is happening is that because you are accessing the route with the internal URL whilst enabling urlPrefixAlias the req url comes appended with the base url. 

If the urlPrefixAlias was http://www.example.com/new_base/ and you accessed the API via http://www.example.com/new_base/resource/paramsAndStuff the req.url val should just be /resource/paramsAndStuff. Instead by using the same actual url as the mask url you are still access the URL by the internal url and the req.url val is being loaded as /rest/example-resource/etc/etc which doubles onto http://localhost:16006/rest/.

What I did, albeit a little inelegant is looks to replace the base param from _apiConfig if urlPrefixAlias is in place.

theninj4:
Magic, it's working a treat :+1: Thanks for the contribution! :smile:

st3xupery:
No problem, I think this is a great project and am excited to be using it in what should shortly be a production environment.

PR #192, opened 2016-09-19
Dynamic swagger.json based on handler functionality

### LINKED ISSUE #190: Swagger Generator should ignore disabled resource options in handlers
You can disable certain resource options in a handler, such as disabling create and delete here:

```javascript

module.exports = new jsonApi.MemoryHandler();
module.exports.create = null;
module.exports.delete = null;
```

This works as expected in the API but unfortunately the swagger generator still seems to generate documentation for these endpoints despite being disabled in an adapter. The swagger generator will need to check for the existence of a function before it generates documentation.

theninj4:
This resolves #190. The example server uses a handler without a `delete` function. The corresponding swagger.json should accurately reflect this.

Starting the example server and browsing to http://localhost:16006/rest/swagger.json then searching for the `/photos/{id}` path should yield just the `get` and `patch` functionality.

screen shot 2016-09-19 at 15 08 22

### DIFF
--- a/lib/swagger/paths.js
+++ b/lib/swagger/paths.js
@@ -38,85 +38,113 @@ swaggerPaths._addPathDefinition = function(paths, resourceConfig) {
 };
 
 swaggerPaths._addBasicPaths = function(paths, resourceName, resourceConfig) {
-  paths["/" + resourceName] = {
-    get: swaggerPaths._getPathOperationObject({
+  var genericPaths = { };
+  var specificPaths = { };
+  paths["/" + resourceName] = genericPaths;
+  paths["/" + resourceName + "/{id}"] = specificPaths;
+
+  if (resourceConfig.handlers.search) {
+    genericPaths.get = swaggerPaths._getPathOperationObject({
       handler: "search",
       resourceName: resourceName,
       description: "Search for " + resourceName,
       parameters: resourceConfig.searchParams,
       hasPathId: false
-    }),
-    post: swaggerPaths._getPathOperationObject({
+    });
+  }
+
+  if (resourceConfig.handlers.create) {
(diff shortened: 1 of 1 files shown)


lib/swagger/paths.js
```diff
@@ -38,85 +38,113 @@ swaggerPaths._addPathDefinition = function(paths, resourceConfig) {
...
       resourceName: relation,
       relationType: relationType,
       extraTags: resourceName,
       hasPathId: true
-    }),
-    post: swaggerPaths._getPathOperationObject({
+    });
+  }
+
+  if (resourceConfig.handlers.update) {
```
pmcnr-hx: `resourceConfig.handlers.create`

lib/swagger/paths.js
```diff
@@ -38,85 +38,113 @@ swaggerPaths._addPathDefinition = function(paths, resourceConfig) {
...
       resourceName: relation,
       relationType: relationType,
       extraTags: resourceName,
       hasPathId: true
-    }),
-    delete: swaggerPaths._getPathOperationObject({
+    });
+  }
+
+  if (resourceConfig.handlers.update) {
```
pmcnr-hx: `resourceConfig.handlers.delete`

lib/swagger/paths.js
theninj4: This is meant to be an `update` - the path handles the creation of a relationship on a resource, which is actually an `update`

lib/swagger/paths.js
theninj4: As above, this is meant to be an `update`

pmcnr-hx:
Good to go: 👍. It will be great to have the more accurate OpenAPI (aka Swagger) docs.

theninj4:
This resolves some of #148

PR #194, opened 2016-09-27
Enable GraphQL

theninj4:
Drop a GraphQL service alongside the JSON:API endpoints. This gives us the choice of consuming our APIs in either format. It's enabled in the example server:

```
$ npm start
```
Then browse to http://localhost:16006/rest/


lib/graphQl/resolvers.js
```diff
@@ -0,0 +1,124 @@
...
+    .then(function(originalResource) {
+      resource = originalResource;
+      return resolvers.rerouteTo("DELETE", path, { }, req, ast);
+    }).then(function() {
+      return resource;
+    });
+};
+
+resolvers.rerouteTo = function(method, path, args, req, ast) {
+  return new Promise(function(resolve, reject) {
```
pmcnr-hx: Warming up to promises? 😉

lib/graphQl/resolvers.js
```diff
@@ -0,0 +1,124 @@
...
+
+resolvers.generateFieldsQueryFromAst = function(ast) {
+  var arrays = (ast.fieldASTs || []).map(function(fieldAST) {
+    return fieldAST.selectionSet.selections || [ ];
+  });
+  var combined = [].concat.apply([], arrays);
+  var fields = []
+  combined.forEach(function(thing) {
+    fields.push(thing.name.value);
+  });
```
pmcnr-hx: Could use a `Array.map` here instead of the `.forEach + .push`, no? (same below)

test/graphQl.js
```diff
@@ -0,0 +1,175 @@
...
+  transport: new Transport('http://localhost:16006/graphql')
+});
+
+
+describe("Testing jsonapi-server graphql", function() {
+
+  describe("read operations", function() {
+
+    it("filter with primary join and filter", function() {
+      return client.query(`
```
pmcnr-hx: ❤️ the great use of template literals.

README.md
```diff
@@ -81,7 +86,7 @@ jsonApi.define({
 
 jsonApi.start();
 ```
-Your new API will be alive at `http://localhost:16006/` and your `photos` resources will be at `http://localhost:16006/photos`.
+Your new API will be alive at `http://localhost:16006/` and your `photos` resources will be at `http://localhost:16006/photos`. The GraphiQL interface will be available at `http://localhost:16006/graphql`.
```
theninj4: It's a thing, if you check the URL you'll be met by an *i*nteractive webapp :+1:

lib/graphQl/resolvers.js
theninj4: Sadly there's no choice :cry:

lib/graphQl/resolvers.js
theninj4: Yes for this one, no for the second one below - that one uses an object instead of an array.

test/graphQl.js
theninj4: The template literals and promises mean we no longer support node v0.10, so this work will have have to form a part of `v2.0.0`. I'll take the time to bump all our dependencies (notably Joi, which has a dependency on later versions of node) and sort out the linting at the same time.

theninj4:
I'm wondering if, instead of having config to control the path of the GraphQL service, we simply use the root of the JSON:API service (which is a 404 right now)?

pmcnr-hx:
I'm a big believer in _convention over configuration_ so I can only support that ☝️.

connormeredith:
Might be an opportunity to get this on http://graphql.org/code/ 😉

pmcnr-hx:
👍 for restricting identifiers to `_`.

lib/graphQl/resolvers.js
```diff
@@ -0,0 +1,123 @@
...
+};
+var rerouter = require('../rerouter.js');
+var jsonApi = require('../..');
+
+
+resolvers.search = function(resourceConfig, attribute, parent, args, req, ast) {
+  var path;
+  // If we don't have a JSON:API resource, go get it
+  if (!parent) {
+    path = jsonApi._apiConfig.pathPrefix + resourceConfig.resource;
```
connormeredith: You can start using `let` here if you like

theninj4: I've got another PR lined up to re-lint the codebase and consider that type of thing :+1:

PR #195, opened 2016-09-28
Relint the project to use eslint:standard

theninj4:
I've moved us away from custom eslint rules to use `standard`, nearly all of these changes are a result of `$ eslint --fix ./lib`, etc. If the build passes we can assume all is good :+1:


connormeredith:
👍  When tests pass

PR #196, opened 2016-09-28
Use ES6 features

theninj4:
Convert the codebase to use all the new ES6 features. Automated via:
```
$ lebab --transform arrow,for-of,arg-spread,obj-method,no-strict,obj-shorthand,let,template,default-param --replace ./lib/
$ eslint --fix ./lib/
```
Again, if the build passes we should be good :+1:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
   matrix:
     - TEST_STEP=jscpd
     - TEST_STEP=lint
-    - TEST_STEP=flow
     - TEST_STEP=test
     - TEST_STEP=swagger
 notifications:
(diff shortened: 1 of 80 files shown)


connormeredith:
👍

theninj4:
Hmm so Node v4 errors with:

theninj4:
So it looks like `node --use_strict` doesn't work with Mocha, so I'm now using `require('use-strict')` in the example server.

theninj4:
That was a no-go either, now I'm trying `mocha --require use-strict`...

theninj4:
Ahhh now Node v4 doesn't support array-spread. I'll close this and trim it a little.

PR #197, opened 2016-09-28
Using es6 features

theninj4:
Convert the codebase to use all the new ES6 features. Automated via:
```
$ lebab --transform arrow,for-of,obj-method,no-strict,obj-shorthand,let,template,default-param --replace ./lib/
$ eslint --fix ./lib/
```
Again, if the build passes we should be good :+1:


connormeredith:
👍  when the build passes

PR #198, opened 2016-09-28
Dependency bump

theninj4:
Updating all our dependencies to the latest versions. Includes a big bump to Joi, whereby we used to be locked in to an old version due to support for Node pre-v4.


connormeredith:
👍  when le tests pass

PR #205, opened 2016-09-29
Enable the injection of an Express Router

theninj4:
This PR allows us to inject an Express Router, simplifying the process of including this project alongside an existing express app.


pmcnr-hx:
👍 once CI passes (seems to be failing for Node.js 4 because strict mode is not enabled).

theninj4:
Doh! I'll put use-strict back in then :(

PR #207, opened 2016-10-21
Added array Joi type for GraphQL

adamlc:
I've added support for arrays with a single scalar type in GraphQL, i.e an array of strings, numbers etc. Although the best practice is to probably have a separate collection sometimes this isn't possible. The spec allows arrays as attributes:

> Complex data structures involving JSON objects and arrays are allowed as attribute values. However, any object that constitutes or is contained in an attribute MUST NOT contain a relationships or links member, as those members are reserved by this specification for future use.

I've also added an embedded tags example to photos and updated the graphql tests :)

### DIFF
--- a/example/resources/photos.js
+++ b/example/resources/photos.js
@@ -24,6 +24,8 @@ jsonApi.define({
       .default(false)
       .description('File in RAW format')
       .example(false),
+    tags: jsonApi.Joi.array().items(jsonApi.Joi.string())
+      .description('Tags for the photo'),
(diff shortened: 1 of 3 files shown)


theninj4:
This is great, sorry I've not reviewed this yet, I'll get to it soon, I promise!

theninj4:
I've got an open question about what happens to the Swagger docs with the Array type, and there's a rogue `console.log`, but this looks good :+1: I'll merge this and pick up the Swagger stuff along with the polymorphic relationships in #213

PR #216, opened 2016-11-08
Dynamodb

anativhq:
you've checked in your keys...

didn't mean to create a PR, sorry!


jokeyrhyme:
Yep, revoke your keys immediately. Assume that bad people have them and are using them to eat up your AWS credits and steal your data.

It might also be worth breaking the DynamoDB handler out into a separate project, similar to the way the SQL and MongoDB handlers have been.

theninj4:
It's cool, these aren't legit keys. I was running a local instance via:
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html#DynamoDBLocal.DownloadingAndRunning

PR #223, opened 2016-12-07
`jsonApi.close()` does not throw for BYO router

### LINKED ISSUE #220: bring-your-own router breaks `jsonApi.close()`
According to the documentation, we can bring our own Express router:
-   https://github.com/holidayextras/jsonapi-server/blob/v2.2.0/documentation/configuring.md

However, this means that the framework no longer has a reference to the HTTP server in use, which means `jsonApi.close()` throws an exception. This is caused when `jsonApi.close()` calls the internal `router.close()` method, which then tries to call `server.close()`:
-   https://github.com/holidayextras/jsonapi-server/blob/v2.2.0/lib/router.js#L98

I have a FaaS use case (as per #219), where I've started a service on a random port, attached this framework to is, and then I want to close both the framework and the service afterwards.

jokeyrhyme:
### Fixed

-   calling `jsonApi.close()` when configured with an existing Express router no longer throws an error (fixes #220)

### Notes

-   I added a test case that configures a JSON API server with an existing Express router, and launches this server in a child process

-   the child process approach seemed to be an easy way to have this additional server run without conflicting with the existing example server

### DIFF
--- a/test/close-byo-router.js
+++ b/test/close-byo-router.js
@@ -1,0 +1,48 @@
+const assert = require('assert')
+const childProcess = require('child_process')
+const path = require('path')
+
+describe('Testing jsonapi-server with bring-your-own router', () => {
+  const serverPath = path.join(__dirname, 'fixtures', 'server-byo-router.js')
+
+  let child
+  let isErrorEmitted
+  let isExitEmitted
+  let exitCode
+  let exitSignal
+
+  it('"exit" event, no "error" event', (done) => {
+    setTimeout(() => {
+      assert.equal(isExitEmitted, true)
+      assert.equal(isErrorEmitted, false)
+      done()
+    }, 1000)
+  })
+
+  it('exit code is 0 (success)', () => {
+    assert.equal(exitCode, 0)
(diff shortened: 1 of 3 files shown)


theninj4:
Great work, Loving the test :+1:

test/close-byo-router.js
```diff
@@ -0,0 +1,48 @@
...
+  before(() => {
+    child = null
+    isErrorEmitted = false
+    isExitEmitted = false
+    exitCode = null
+
+    child = childProcess.fork(serverPath, [], { stdio: 'inherit' })
+    child.on('error', () => { isErrorEmitted = true })
+    child.on('exit', (code, signal) => {
+      exitCode = code
```
theninj4: I hit a timing issue on Travis when cutting a new release, I ended up making the `before` block async and letting the test progress once the child has exited:
https://github.com/holidayextras/jsonapi-server/pull/226/commits/9e5111ec369427d86b3e3086d867af3ff22f0477
example build failure (only in Node v4, only some of the time):
https://travis-ci.org/holidayextras/jsonapi-server/builds/182226747

PR #224, opened 2016-12-07
add docs and ChainHandler support for handler `close()`

### LINKED ISSUE #221: should `handlers.close()` be officially documented and encouraged?
I found that `jsonApi.close()` calls `handlers.close()`:
-   https://github.com/holidayextras/jsonapi-server/blob/v2.2.0/lib/jsonApi.js#L126

But this doesn't seem to be an officially documented part of the handler specification:
-   https://github.com/holidayextras/jsonapi-server/blob/v2.2.0/documentation/handlers.md

Is it worth adding this to the official handler specification?
Is implementing a `.close()` method something that we should encourage?

jokeyrhyme:
### Added

-   document the optional `close()` method that handlers may implement (fixes #221)

-   `ChainHandler` supports the optional handler `close()` method if implemented

### Changed

-   `ChainHandler` supports handlers that do not implement `initialise()`

### DIFF
--- a/documentation/handlers.md
+++ b/documentation/handlers.md
@@ -6,6 +6,7 @@ Handlers represent the mechanism that backs a resource. Each handler is an objec
 * a `ready` property indicating the handler is ready to process requests.
 * some of the following methods:
  * `initialise` - when jsonapi-server loads, this is invoked once for every resource using this handler. Its an opportunity to allocate memory, connect to databases, etc.
+ * `close` - for cleaning up upon `jsonApi.close()` (optional)
  * `search` - for searching for resources that match some vague parameters.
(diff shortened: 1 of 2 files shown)


jokeyrhyme:
This has the unintended side-effect of preventing `lib/rerouter.js` from being used for network-free requests.

Eventually, I'd like to be able to do something like:

```js
jsonApi.start()
// TODO: use lib/rerouter.js or other network-free request approach
jsonApi.close()
```

If we merge this, then we probably need a new option for `jsonApi.setConfig()` in order to allow `lib/rerouter.js` to still work?

theninj4:
:+1: Looks great, thanks for adding to the documentation! :)

PR #225, opened 2016-12-07
prefer client-generated ID if provided

### LINKED ISSUE #218: support for client-generated IDs ?
I just found this JSON API implementation and it's going to save me a tonne a time. Thanks!

I'm trying to use this with client-generated IDs: http://jsonapi.org/format/#crud-creating-client-ids

My use case is that I'm trying to create deterministic automated tests, and this would be easier if I could just hardcode the IDs for any resources that the tests will be manipulating.

I've tried POSTing resources with client-generated IDs with both the memory and MongoDB handlers, and I seem to be getting server-generated IDs all the time.

There are ways to work around this, so I'm not roadblocked or about to throw a tantrum or anything. :) Admittedly, outside of testing, I am unlikely to find this especially handy.

Given that the upstream JSON API specification says that servers "MAY" implement this support, is there any desire or intention to do so within this project?

jokeyrhyme:
### Changed

-   prefer client-generated ID (if provided) when creating new resources (fixes #218)

### DIFF
--- a/lib/routes/create.js
+++ b/lib/routes/create.js
@@ -32,10 +32,15 @@ createRoute.register = () => {
       },
       callback => {
         const theirs = request.params.data
-        theirResource = _.assign({
-          id: uuid.v4(),
-          type: request.params.type
-        }, theirs.attributes, { meta: theirs.meta })
+        theirResource = _.assign(
+          {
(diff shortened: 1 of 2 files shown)


theninj4:
:+1: looks great :)

PR #228, opened 2016-12-15
Check resource schema are valid on startup

### LINKED ISSUE #201: Process resources on startup to validate relationships are valid
If the resource configuration is invalid, we should error as early as possible, and provide as much information as possible.

For example if we have two resources `A` and `B`, whereby `A` has many `B` and `B` belongs to many `a`, we should error on startup to say `Resource B belongs to many 'a', but 'a' isn't a valid resource`.

theninj4:
Resolves #201. Once `myServer.start()` has been invoked, we walk through all the resource definitions ensuring they're all valid.

If a resource has been miss-spelt when defining a relation, you'll get an error like this
```
Error: Resource 'comments' lists attribute 'article' as having a relation with 'article', but 'article' is not a valid resource name!
```

If a foreign relation has been miss-spelt, you'll get an error like this:
```
Error: 'comments'.'article' is defined as being a foreign relation to the primary 'articles'.'comment', but that primary relationship does not exist!
```

### DIFF
--- a/lib/schemaValidator.js
+++ b/lib/schemaValidator.js
@@ -10,7 +10,7 @@ schemaValidator.validate = function (resources) {
       let types = joiSchema._settings.__one || joiSchema._settings.__many
       types.forEach(type => {
         if (!resources[type]) {
-          throw new Error(`Error: '${resource}'.'${attribute}' is defined to hold a relation with '${type}', but '${type}' is not a valid resource name!`)
+          throw new Error(`'${resource}'.'${attribute}' is defined to hold a relation with '${type}', but '${type}' is not a valid resource name!`)
         }
       })
       let foreignRelation = joiSchema._settings.__as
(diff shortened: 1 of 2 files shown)


lib/schemaValidator.js
```diff
@@ -0,0 +1,25 @@
...
+      let types = joiSchema._settings.__one || joiSchema._settings.__many
+      types.forEach(type => {
+        if (!resources[type]) {
+          throw new Error(`'${resource}'.'${attribute}' is defined to hold a relation with '${type}', but '${type}' is not a valid resource name!`)
+        }
+      })
+      let foreignRelation = joiSchema._settings.__as
+      if (!foreignRelation) return
+
+      let backReference = resources[types[0]].attributes[foreignRelation]
```
pmcnr-hx: Since all the logic here is read-only, may change all the `let`s above to `const`s to make explicit at a glance that there's no mutation of those variables occurring.

theninj4: Good call :+1:

PR #233, opened 2016-12-27
Multiple filter w/ include fix + internal routing fix + more

st3xupery:
Applied a fix that was breaking the ability to acquire includes when using multiple filters through a comma separated list. It seems I was splitting the filter param by comma too far into a particular function that wasn't called when internally routing includes. You'll see the filter split happens immediately in the parseandValidate function. I wrote up two tests for this and added an additional photo resource which required me to modify the counts in a few test assertions.

Additionally in the include lib when building a list of resources to fetch I reduced the url to just the base + path of the url as the hostname is not required while routing internally. This also remedies a problem where the use of pathPrefix which is overwritten by the apiPrefixOverride would cause this route to be incorrect from an internal standpoint.

Finally I appended originalUrl to the request object which persists as the same value during rerouting as it allows me to compare the route object to the originalUrl to see if the request is the result of an include or whatever else may prompt a reroute. I feel like this is a quasi acceptable use of the param as defined in the express API here http://expressjs.com/en/api.html#req.originalUrl

### DIFF
--- a/lib/filter.js
+++ b/lib/filter.js
@@ -67,8 +67,6 @@ filter._parseScalarFilterElement = (attributeConfig, scalarElement) => {
 filter._parseFilterElementHelper = (attributeConfig, filterElement) => {
   if (!filterElement) return { error: 'invalid or empty filter element' }
 
-  if (typeof filterElement === 'string') filterElement = filterElement.split(FILTER_SEPERATOR)
-
   const parsedElements = [].concat(filterElement).map(scalarElement => filter._parseScalarFilterElement(attributeConfig, scalarElement))
 
   if (parsedElements.length === 1) return parsedElements[0]
@@ -116,6 +114,8 @@ filter.parseAndValidate = request => {
 
   for (const key in request.params.filter) {
     filterElement = request.params.filter[key]
+
+    if (typeof filterElement === 'string') request.params.filter[key] = filterElement = filterElement.split(FILTER_SEPERATOR)
 
     if (!Array.isArray(filterElement) && filterElement instanceof Object) continue  // skip deep filters
 
--- a/lib/graphQl/index.js
+++ b/lib/graphQl/index.js
@@ -13,10 +13,12 @@ const writeTypes = require('./writeTypes.js')
 jsonApiGraphQL.with = app => {
   const config = jsonApi._apiConfig
 
-  app.use(new RegExp(`${config.base}$`), graphqlHTTP({
-    schema: jsonApiGraphQL.generate(jsonApi._resources),
-    graphiql: !!config.graphiql
-  }))
+  if (config.graphiql !== false) {
+    app.use(new RegExp(`${config.base}$`), graphqlHTTP({
+      schema: jsonApiGraphQL.generate(jsonApi._resources),
+      graphiql: !!config.graphiql
+    }))
+  }
 }
 
 jsonApiGraphQL.generate = allResourceConfig => {
--- a/lib/swagger/index.js
+++ b/lib/swagger/index.js
@@ -17,8 +17,8 @@ swagger._getSwaggerBase = () => {
   let basePath, host, protocol
   if (jsonApi._apiConfig.urlPrefixAlias) {
     const urlObj = url.parse(jsonApi._apiConfig.urlPrefixAlias)
-    basePath = urlObj.pathname.replace(/\/$/, '')
-    if (basePath.length === 0) basePath = '/';
+    basePath = urlObj.pathname
+    if (basePath !== '/') basePath = basePath.replace(/\/$/, '')
     host = urlObj.host
     protocol = urlObj.protocol.replace(/:$/, '')
   } else {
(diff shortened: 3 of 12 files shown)


st3xupery:
There was also an issue where setting graphiql as false would still run the joiconverter process to build the definition. So now if the param is false it skips executing the process.

The API I'm building has attributes that are objects which throws errors with the Joi to GraphQL converter so maybe it's something to be fixed, but in the interim I'd just like to have it off.

theninj4:
@pmcnr-hx would you like to take a review on this one? You were the closest to the filter preprocessing a few months back.

I'm fine with the idea of turning off all of GraphQL with a strict `false` in the config :+1:

pmcnr-hx:
I'll be reviewing this soon, but the changes are not trivial, so I'll need to take some time.

st3xupery:
Is there anything I can do to help you along?

pmcnr-hx:
I'll review today @st3xupery. Meanwhile can you merge or rebase on master?

st3xupery:
@pmcnr-hx By that what do you mean. Do you want me to make my pull request off of the master of my fork? I'm still using the master for legacy support of one of my projects. I can make another fork if need be.

pmcnr-hx:
@st3xupery. What you just did. Thanks. :) I'm moving this to the top of my TODO list. Should be able to get round to it soon. Apologies for the delay.

st3xupery:
@pmcnr-hx Okay, great and no worries~!

lib/filter.js
```diff
@@ -117,6 +115,8 @@ filter.parseAndValidate = request => {
   for (const key in request.params.filter) {
     filterElement = request.params.filter[key]
 
+    if (typeof filterElement === 'string') request.params.filter[key] = filterElement = filterElement.split(FILTER_SEPERATOR)
```
pmcnr-hx: This is a pragmatic solution, but I'm not too thrilled about mutating `request.params.filter[key]`. I'm going to have a look at the internal routing and check if we can used the processed filter there also.

lib/graphQl/index.js
```diff
@@ -13,10 +13,12 @@ const writeTypes = require('./writeTypes.js')
 jsonApiGraphQL.with = app => {
   const config = jsonApi._apiConfig
 
-  app.use(new RegExp(`${config.base}$`), graphqlHTTP({
-    schema: jsonApiGraphQL.generate(jsonApi._resources),
-    graphiql: !!config.graphiql
-  }))
+  if (config.graphiql !== false) {
```
pmcnr-hx: Shall we introduce the `strict` configuration property that @theninj4 [suggests](https://github.com/holidayextras/jsonapi-server/pull/233#issuecomment-269498049) to turn off the GraphQL endpoint? Abusing the `graphiql` configuration property for this end is not the best option.

lib/swagger/index.js
```diff
@@ -17,7 +17,8 @@ swagger._getSwaggerBase = () => {
   let basePath, host, protocol
   if (jsonApi._apiConfig.urlPrefixAlias) {
     const urlObj = url.parse(jsonApi._apiConfig.urlPrefixAlias)
-    basePath = urlObj.pathname.replace(/\/$/, '')
```
pmcnr-hx: If we use the `(?<!^)\/$` regexp (with a look behind to assert we're not at the beginning of the string) above we can avoid the `if`.

yossialush:
Hi, this fix is kind of crucial for us. It's blocking us from upgrading to latest version, thus we stuck with old bugs. When should we anticipate for this merge to take place?

theninj4:
It seems like there are no big objections to what's going on here. As soon as the build goes green, I'll merge and release. Looks like the last commit broke something, it looks unrelated to the rest of this pr, so if the build is still failing tomorrow midday GMT I might merge this PR then revert the troublesome commit.

yossialush:
Good news! Thanks 🎉

st3xupery:
Hey! Sorry I kind of neglected this. 
The last commit was a fix to a whole other issue. I'll revert this branch to the previous commit and make a pull on the other issue separately. 
As far as @pmcnr-hx are concerned. I'm not too clear on the 2nd, isn't technically !== false a strict equivalency, and I can implement the regexp expression in his 3rd suggestion.
I'll do this all by tonight EST.

st3xupery:
I've made the regex replacement of the if statement. Had to use something slightly different as javascript does not have a *look behind* in it's regex capabilities. The alternative works to the same effect: `(?!^\/)\/$`

lib/swagger/index.js
st3xupery: Had to use something slightly different as javascript does not have a look behind in it's regex capabilities. The alternative works to the same effect: (?!^\/)\/$

lib/graphQl/index.js
st3xupery: Not sure what you mean, is !== false not a strict equivalency?

pmcnr-hx: What we were suggesting was a *new* configuration property named `strict`, but we can get round to doing that in a later PR.

PR #238, opened 2016-12-28
Fix bug when computing last page url

### LINKED ISSUE #230: Last page link calculation is not accurate
I've noticed that pagination link for the last page is miscalculated when the total result is even and
the limit is odd. For example: using this pagination params: page[limit]=5, I've got the following result:

"meta": {
    "page": {
      "offset": 0,
      "limit": 5,
      "total": 16
    }
  },
  "links": {
    "self": "/containers/12/templates?page[limit]=5",
    "last": "/containers/12/templates?page%5Blimit%5D=5&page%5Boffset%5D=14",
    "next": "/containers/12/templates?page%5Blimit%5D=5&page%5Boffset%5D=5"
  }

As you can see, last page has offset of 14 instead of 15.

theninj4:
Resolves #230 - The maths for computing the last page is now much better, although not perfect.

Assuming we have 16 resources and a page-limit of 4, the pages should be:
```
0 to 3 (first)
4 to 7
8 to 11
12 to 16 (last)
```
With a page-limit of 5, they should be:
```
0 to 4 (first)
5 to 9
10 to 11
15 to 16 (last)
```
And with a page-limit of 6, they should be:
```
0 to 5 (first)
6 to 11
12 to 16 (last)
```
I've added some unit tests to validate the above.

### DIFF
--- a/test/pagination.js
+++ b/test/pagination.js
@@ -1,6 +1,7 @@
 const assert = require('assert')
 const helpers = require('./helpers.js')
 const jsonApiTestServer = require('../example/server.js')
+const pagination = require('../lib/pagination.js')
 
 let pageLinks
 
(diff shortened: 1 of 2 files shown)


kazazor:
@theninj4 - do you have any estimation on when that will be merged and published in a patch version?

By the way, thank you for your fix :)

theninj4:
@kazazor good call, thanks for the reminder :+1: I'll chase up a review

test/pagination.js
```diff
@@ -125,31 +126,36 @@ describe('Testing jsonapi-server', () => {
       })
     })
 
-    it('fetches an obscure page', done => {
-      const data = {
-        method: 'get',
-        url: 'http://localhost:16006/rest/articles?page[offset]=1&page[limit]=2&sort=title'
+    describe.only('correctly computes the last pages', done => {
```
pmcnr-hx: Oops... The `.only` needs to go.

theninj4:
@pmcnr-hx resolved :+1:

pmcnr-hx:
Good to go! :+1:

PR #241, opened 2017-01-05
start() and close() return promises

shaunc:
This is useful (e.g. in testing) when you want to wait to do something once the server is up, or take other cleanup actions once the server is down.

### DIFF
--- a/lib/jsonApi.js
+++ b/lib/jsonApi.js
@@ -116,8 +116,10 @@ jsonApi.start = () => {
   router.applyMiddleware()
   routes.register()
   if (!jsonApi._apiConfig.router) {
-    router.listen(jsonApi._apiConfig.port)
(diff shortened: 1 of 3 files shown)


shaunc:
Interested? Want me to rebase?

theninj4:
I love the idea :+1: These projects favour callbacks over promises, and and use things like `Promise.denodeify` to enable support for both when exposing functionality outwards. I'd rather keep this codebase promise-free if at all possible.

shaunc:
Ah ... jsonapi-client uses promises, so I thought they would be ok here too...

"Promise.denodify": ie support convention "for async support, pass in a callback; if no callback is passed in a promise is returned"?

Would it be ok to support both here? I can modify the PR accordingly if so.

PR #262, opened 2017-03-05
Added descriptions to deep relationships in swagger.json

st3xupery:
Kind of trivial, but I hope the added description could be instructive that sees these deep relationship URIs and gets a slightly better sense of what they are for.

Correct me if any of my descriptions are incorrect.


paparomeo:
This looks reasonable and useful. I'll take a better look to make sure the descriptions are correct and then merge.

PR #264, opened 2017-03-28
Fix bug when specifying include and fields.

wolfgang42:
I discovered a bug when specifying `fields` with included resources: the `fields` is applied to *all* resources, meaning that other resources also had their attributes stripped. This PR adds test cases for and fixes this bug. I also suggest disallowing `var`, which would have caught this bug to begin with.

See the commit messages for further exposition.


phil-mellor:
Thanks for this - I am experiencing the same issue and this change fixes it.

Issue #263, opened 2017-03-19
package.json requires eslint-plugin-promise twice


flyingbuddha:
Hello 👋,

`eslint-plugin-promise` is defined twice in package.json with conflicting version numbers.
In dependencies it's listed as **^3.3.0**, whilst in devDependencies it's listed as **^2.0.0**.

What is the correct course of action? I feel that it should be removed from dependencies and the version bumped to **^3.3.0** in devDependencies.

pmcnr-hx:
Apologies for the late reply @flyingbuddha. Just merged @jmeas's kindly contributed pull request to fix this.

PR #266, opened 2017-04-11
Remove eslint plugin from dependencies

jmeas:
It appears that this may have accidentally been added as a dependency in ddc26f0e3428

Resolves gh-263


jmeas:
Thanks for the merge, @pmcnr-hx !

PR #272, opened 2017-04-19
Properly merge arrays of filters when merging params

slickmb:
* fixes issues when filters are added to relationships links, where the filters are not merged properly when being rerouted.  More specifically, this fails when there are multiple filters on the same field (e.g. `/type/id/relation?filter[created_at]=>2016-01-01&filter[created_at]=<2017-01-01`)


honzas:
Hi, any idea if this will be merged any time soon? This fix is far more important than just some merging of filters.. 

If you want to fetch multiple types in one relation in graphql, only request for first type succeeds. All other have broken request because of the current implementation of _mergeParams (it breaks second parameter which is then used in following requests)

In other words it should fix this kind of graphql query (which is currently broken)

```
{
  listing {
    listItems {
      __typename
      ... on apple{
        id
      }
      ... on orange{
        id
        title
      }
    }
  }
}
```

paparomeo:
I have time set aside next week for the project and will review and merge this @honzas. Thank you for your report that this fixes your GraphQL use case.

championswimmer:
@slickmb shouldn't the operator be `>=` and not `=>` (same for less)

slickmb:
cleaning up old PRs

PR #301, opened 2017-06-28
add test that shows graphQl varaibles do not work

laurieb:
Hi,

I created this PR to demonstrate the issue I'm having. 

Currently the GraphQl implementation does not support query variables as far as I can get it to work.

I've added a basic test that uses a named query with a variable passed in a manner which is consistent with the [Lokka](https://github.com/kadirahq/lokka) documentation.

I don't think the GraphQl server parses out the variables properly.

Am I wrong or do you think I might be on to something?

Great work on the server BTW!

Thanks
Laurie

### DIFF
--- a/test/graphQl.js
+++ b/test/graphQl.js
@@ -41,7 +41,33 @@ describe('Testing jsonapi-server graphql', () => {
       })
     }))
 
-    it('filter with foreign join and filter', () => client.query(`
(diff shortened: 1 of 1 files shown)


laurieb:
I've tried it through the GraphiQl interface as well with no luck...
screen shot 2017-06-28 at 18 23 09

PR #306, opened 2017-07-14
Apply use strict to missing files

joseph-norman:
Apply `"use strict"` to the missing files in the `jsonapi-server` to ensure that all libraries utilising the package as a dev dependency (for unit tests, etc). Can run them fine when using block-level scope variables.


joseph-norman:
lints 🔨

PR #307, opened 2017-07-14
Applying temporary forced order to graphql tuple test

joseph-norman:


test/polymorphicRelationships.js
```diff
@@ -117,7 +117,18 @@ describe('Testing jsonapi-server', () => {
         }
       }
     `).then(result => {
-      assert.deepEqual(result, {
+      /**
+       * This is a pragmatic fix to guarantee order. Tuples in GraphQL
```
pmcnr-hx: Can you add a `FIXME` here, so we can track it more easily?

joseph-norman: 👍  Will do!

PR #320, opened 2017-09-28
Add resourceConfig.generateId to disable automatic ID generation.

### LINKED ISSUE #259: Add resource configuration to disable automatic ID generation
I have some resources whose IDs are integers auto-generated by the database server. I declare them as:
```javascript
jsonApi.define({
    resource: 'Order',
    attributes: {
        id: jsonApi.Joi.number().integer(),
        // Other attributes
    },
    handler: customHandler,
})
```

This causes a problem when I try to create a new ID: a UUID is automatically generated, but then fails validation because it is not an integer.
I therefore propose a new key in the resource definition, `generateId`, which can be set to `false` to disable autogenerating IDs. I have implemented this on my branch [create-generate-id](/holidayextras/jsonapi-server/compare/master...wolfgang42:create-generate-id), and it is working well for my use so far.
In the future this key could also be extended to take a function which returns an ID, so a project could define its own ID generation scheme. (For example using [`shortid`](https://www.npmjs.com/package/shortid); or perhaps even a content-addressable system where a data property is hashed to generate the ID.)

wolfgang42:
Closes #259.


paparomeo:
I'm open to merging this, since there are a number of arguments for it. However, I will need some test coverage to go with the code changes.

wolfgang42:
Of course! I made this change a long time ago and had completely forgotten that it was missing tests. I'll add them in the next couple of days and update the PR.

wolfgang42:
Added tests & documentation.

championswimmer:
@wolfgang42 
I have started maintaining a fork of this https://github.com/coding-blocks/jsonapi-server with primarily the concern of using integer autoincrement ids in my mind. The "id" field is hardcoded all over the place in this implementation, so I did a complete overhaul and changed all mentions to number instead of string/uuid.
 You might want to take a look at it.

wolfgang42:
@championswimmer I considered overriding the data type, but the [JSON-API spec says](http://jsonapi.org/format/#document-resource-object-identification):
> The values of the `id` and `type` members **MUST** be strings.

In the interests of maintaining a properly conformant API I therefore cast them to strings when they come out of the database and the rest of the code treats them as such; when I put them back in the server handles the casting automatically. Since they're IDs everything else just treats them as opaque anyway.

paparomeo:
@wolfgang42 is correct. I should have time this week to merge this.

championswimmer:
@paparomeo @pmcnr-hx @wolfgang42 

I have added support for generateId flag in relationaldb too 
https://github.com/holidayextras/jsonapi-store-relationaldb/pull/90

PR #328, opened 2017-10-23
[DRAFT] use integer ( maybe autoincrement on SQL) ids instead of UUID

championswimmer:
this is not a "ready to merge" PR. 

I took at look at #320 and it is not adequate to properly support integer IDs
This PR is the entire kitchen sink we need to be able to properly support integer IDs in **both** graphql and jsonapi.

I need some guidance on how best can this be integrated ? (using a possible resourceConfig.generatedId option)

### DIFF
--- a/lib/routes/create.js
+++ b/lib/routes/create.js
@@ -33,10 +33,8 @@ createRoute.register = () => {
       callback => {
         const theirs = request.params.data
         theirResource = _.assign(
-          {
-            id: 0,
-            type: request.params.type
-          },
+          { type: request.params.type },
+          (request.resourceConfig.generateId !== false) && { id: uuid.v4() },
           theirs.id && { id: theirs.id },
           theirs.attributes,
           { meta: theirs.meta }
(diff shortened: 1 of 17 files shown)


lib/routes/create.js
```diff
@@ -34,7 +34,7 @@ createRoute.register = () => {
         const theirs = request.params.data
         theirResource = _.assign(
           {
-            id: uuid.v4(),
+            id: 0,
```
championswimmer: This is 0 here as a "DEFAULT" value.
Putting the onus on the store handler (relattionaldb) to treat 0 = DEFAULT and if the value is 0, then generate new id, if not 0, then use it like a client provided id. 

see https://github.com/championswimmer/jsonapi-store-sequelize/commit/ee0b9f0303f9101577807e0dd8b720b8f67ab9d7#diff-8004fa36f21fa5eeb3c49b89106f6942R9 for example implementation

pmcnr-hx:
Now that we've merged #320, and as explained there the spec requires the id to be a string, do we still need this PR @championswimmer, or can we close it?

PR #332, opened 2017-10-28
Breaking change: Return error when a response item does not validate.

### LINKED ISSUE #330: Why do schema validation errors not return an error to the client?
If a handler returns an invalid object, it is silently discarded by `responseHelper._enforceSchemaOnObject()`, with the only warning being a debug message. This causes a great deal of confusion when items silently go missing from responses. Is there a reason that it's set up this way? It makes much more sense to me that this would cause an error, something like:
```javascript
return callback({
  status: 500,
  code: 'EINVALIDITEM',
  title: 'Item in response does not validate',
  detail: {
    item: item,
    error: err.message
  }
})
```

I'd submit a pull request, but I wanted to ask first if this should be behind an option in case someone is relying on this behaviour.

wolfgang42:
Previously, the item would be silently dropped with only a debug message.

This is a breaking change because it changes the behaviour of responses for existing handlers which return an item that does not validate.

Closes #330.

### DIFF
--- a/example/handlers/brokenResponseHandler.js
+++ b/example/handlers/brokenResponseHandler.js
@@ -1,0 +1,5 @@
+'use strict'
+
+const jsonApi = require('../..')
+
+module.exports = new jsonApi.MemoryHandler()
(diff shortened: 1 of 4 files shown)


joseph-norman:
Thanks for the contribution @wolfgang42 - looks great! Merging.

PR #362, opened 2018-01-19
Remove Url.parse limit of parameters (default 1000)

talanta:
When you query more than 1000 items including relationships, you will get only 1000 elements in the `included` array. This is due to the default limit of the `url.parse` method from `qs` (https://github.com/ljharb/qs/blob/master/README.md#parsing-objects)

### DIFF
--- a/documentation/configuring.md
+++ b/documentation/configuring.md
@@ -26,7 +26,11 @@ jsonApi.setConfig({
   // (optional) meta can be a function to be invoked at the end of every request
   meta: function(request) {
     return { timestamp: new Date() };
-  }
+  },
+  // (optional) queryStringParsingParameterLimit allows to
+  // override the default limit of 1000 parameters in query string parsing,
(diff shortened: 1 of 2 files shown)


pmcnr-hx:
Thank you very much for the contribution @talanta. I'm happy to make the limit configurable, however, like the documentation notes, it is there for good reason to prevent abuse. Can you update your PR to read the value from the [configuration object](https://github.com/holidayextras/jsonapi-server/blob/master/documentation/configuring.md#configuring-jsonapi-server) instead of hardcoding it to `infinity`?

talanta:
Hi @pmcnr-hx, thank you for the review .. I updated the PR according to your suggestions. Let me know if it's ok.

paparomeo:
Thanks for updating the PR. I should be able to review it again (and hopefully merge) soon.

paparomeo:
Thanks again for the PR @jstafford! Merged and I'll doing a fresh minor release that includes it.

PR #367, opened 2018-01-23
Bug: fix crud operation detection when using chain handlers

slickmb:
Problem: I ran into an issue where I had a ChainHandler pair with a store that did not support all the crud operations.  I was expecting jsonapi to return a 403 action not supported error, but instead got a 500 when the chainHandler attempted to call the action on the underlying store (which failed with a null ref error).

### DIFF
--- a/lib/routes/helper.js
+++ b/lib/routes/helper.js
@@ -71,7 +71,13 @@ helper.verifyRequest = (request, resourceConfig, res, handlerRequest, callback) 
     })
   }
 
-  if (!resourceConfig.handlers[handlerRequest]) {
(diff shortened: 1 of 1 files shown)


paparomeo:
Thank you for the contribution. I should be able to review and merge it soon.

championswimmer:
Just to inform, 
I have used your patch in my fork 
https://github.com/coding-blocks/jsonapi-server/commit/438fe426b7a9f3db6a4d7aa5532719b67525cbd5 

THanks

paparomeo:
Your patch makes a lot a sense. Thanks for the contribution @slickmb. I shall merge and make a release.

PR #382, opened 2018-02-27
Support filter relationships from sparse fieldsets

### LINKED ISSUE #269: Sparse Fieldsets Compliance
When specifying particular fields for a resource, the system seems to ignore filtering relationships, including them all regardless of the client's requests. This does not comply with the specification sheet [here](http://jsonapi.org/format/#fetching-sparse-fieldsets)

> If a client requests a restricted set of fields for a given resource type, an endpoint MUST NOT include additional fields in resource objects of that type in its response.

Additionally, the term fields does include relationships as can be seen [here](http://jsonapi.org/format/#document-resource-object-fields)

> A resource object’s attributes and its relationships are collectively called its “fields”.

wgator:
Should resolve #269.

I'm checking the if the parameter ```removeFilteredRelationships``` from jsonApi config is true to support backward compatibility.

What do you guys think?

### DIFF
--- a/lib/postProcessing/fields.js
+++ b/lib/postProcessing/fields.js
@@ -34,12 +34,25 @@ fields.action = (request, response, callback) => {
     }
   }
 
+  const removeRelationships = jsonApi._apiConfig.removeFilteredRelationships
   allDataItems.forEach(dataItem => {
+    // Removing fields from attributes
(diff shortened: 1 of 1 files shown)


paparomeo:
Thank you very much for the PR @wgator! At a glance it looks good, but I'll be able to review it properly and hopefully merge it next week.

